PR #63, opened 2015-02-06
inotify: code cleanup. Closing watcher should now always shut down goroutine

PieterD:
fixes go-fsnotify/fsnotify#3

Also fixed a racey access to w.watches in Close.

I put the tests in notify_test.go, so it doesn't break any other builds. These tests should be checked against the other platforms one by one before pulling them into integration_test.go.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -68,21 +68,13 @@ func (w *Watcher) Close() error {
 	// Remove all watches.
 	// Everything after this may generate errors because the inotify channel
 	// has been closed; we don't care.
-	w.mu.Lock()
-	hasWatches := false
-	for name := range w.watches {
-		err := w.remove(name)
-		if err == nil {
-			hasWatches = true
-		}
-	}
-	w.mu.Unlock()
+	numWatches := w.removeAll()
 
 	// If no watches were removed, it's possible syscall.Read is still blocking.
 	// In this case, create a watch and remove it to wake it up.
 	// If that fails, there's really nothing left to do. we've done our best,
 	// but the goroutine may be alive forever.
-	if !hasWatches {
+	if numWatches == 0 {
 		wd, _ := syscall.InotifyAddWatch(w.fd, ".", syscall.IN_DELETE_SELF)
 		syscall.InotifyRmWatch(w.fd, uint32(wd))
 	}
@@ -123,17 +115,19 @@ func (w *Watcher) Add(name string) error {
 	return nil
 }
 
-// Remove stops watching the the named file or directory (non-recursively).
+// Remove stops watching the named file or directory (non-recursively).
 func (w *Watcher) Remove(name string) error {
+	name = filepath.Clean(name)
+	if w.isClosed() {
+		return errors.New("inotify instance already closed")
+	}
+
+	// Fetch the watch.
 	w.mu.Lock()
 	defer w.mu.Unlock()
-	return w.remove(name)
-}
-
-// Only call this when w.mu is locked, like from Remove or Close.
-func (w *Watcher) remove(name string) error {
-	name = filepath.Clean(name)
 	watch, ok := w.watches[name]
+
+	// Remove it from inotify.
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
@@ -143,6 +137,21 @@ func (w *Watcher) remove(name string) error {
 	}
 	delete(w.watches, name)
 	return nil
+}
+
+// Called from Close(); remove all watches.
+func (w *Watcher) removeAll() int {
+	successful := 0
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	for name, watch := range w.watches {
+		success, _ := syscall.InotifyRmWatch(w.fd, watch.wd)
+		if success != -1 {
+			successful++
+		}
+		delete(w.watches, name)
+	}
+	return successful
 }
 
 type watch struct {
--- a/inotify.go
+++ b/inotify.go
@@ -19,14 +19,13 @@ import (
 
 // Watcher watches a set of files, delivering events to a channel.
 type Watcher struct {
-	Events   chan Event
-	Errors   chan error
-	mu       sync.Mutex        // Map access
-	fd       int               // File descriptor (as returned by the inotify_init() syscall)
-	watches  map[string]*watch // Map of inotify watches (key: path)
-	paths    map[int]string    // Map of watched paths (key: watch descriptor)
-	done     chan bool         // Channel for sending a "quit message" to the reader goroutine
-	isClosed bool              // Set to true when Close() is first called
+	Events  chan Event
+	Errors  chan error
+	mu      sync.Mutex        // Map access
+	fd      int               // File descriptor (as returned by the inotify_init() syscall)
+	watches map[string]*watch // Map of inotify watches (key: path)
+	paths   map[int]string    // Map of watched paths (key: watch descriptor)
+	done    chan struct{}     // Channel for sending a "quit message" to the reader goroutine
 }
 
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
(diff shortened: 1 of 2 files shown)


nathany:
Thanks.

inotify.go
```diff
@@ -101,9 +125,14 @@ func (w *Watcher) Add(name string) error {
 
 // Remove stops watching the the named file or directory (non-recursively).
 func (w *Watcher) Remove(name string) error {
-	name = filepath.Clean(name)
 	w.mu.Lock()
 	defer w.mu.Unlock()
+	return w.remove(name)
+}
+
+// Only call this when w.mu is locked, like from Remove or Close.
```
nathany: I'm not in love with having a function that requires the mutex to be done outside of it, but I don't have any better ideas right now.

Also not sure about taking a Mutex around the entirety of remove vs. just some part of it.

inotify_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright 2010 The Go Authors. All rights reserved.
```
nathany: © 2015 for new files, old files can rename the same

inotify.go
PieterD: How about a seperate method to return a slice of watches, or a method to simply remove all watches?

Since the watcher is closed when Closing anyway, no new ones will be added, and we could do more fine-grained locking in remove.

inotify_test.go
PieterD: Got it. Sorry.

inotify.go
PieterD: I'm not convinced fine-grained locking is an option here. It seems a trap for race conditions at first glance. Maybe it's even a mistake in Add.

nathany:
Have you confirmed that #3 is fixed by this change?

I'll take a closer look and do some QA tomorrow before I merge it in. Thanks again for the help.

PieterD:
I've checked. I did the same as you (checked /proc/x/fd), disabled tests until i found the one that did it. Unsurprisingly, it was a blocking goroutine.

It still happened due to corner case that I hadn't covered yet; in close, when there were watches to remove but they all gave errors, the 'wakeup' wasn't done. Now it is. And thus it is fixed.

nathany:
@PieterD Have you already signed Google's CLA for contributing to Go? (This code base originates from Go exp packages and may one day land in Go).

https://developers.google.com/open-source/cla/individual

PieterD:
Did that (and A+C) years ago. It's fine.

PieterD:
There's still some uncertainty when closing which I'm fixing now.

nathany:
Thanks. I'm on my way to a hack day, fsnotify will be my project for today.

PieterD:
Nice. I'm using epoll to get rid of the last bit of uncertainty.

nathany:
And then it gets interesting. :-)

PieterD:
Yeah, it's a lot of busywork for a small imperfection, but it's got to be done. You could accept the pull as is, and I'll shoot the epoll thing as a seperate pull request. That might ease the review.

PieterD:
Man, it SUCKS that syscall.Read doesn't unblock with you close the file descriptor. Even epoll doesn't respond when you do that.

nathany:
Does IN_NONBLOCK help at all? See #5.

nathany:
I admit I don't understand this stuff. I'm yet to become a Linux guru. :-)

nathany:
A separate PR would be appreciated. Let me get this tested and merged in.

PieterD:
Please. It would be nice to seperate this mess and the next mess, instead of creating one giant ubermess :P

nathany:
übermess!

PieterD:
HOLY CRAP I just figured it out. There's a truly awesome race condition that happens :D

I was wondering how it could be possible that sometimes, test n was making test n+1 fail with "bad file descriptor" or with the wrong number of events.

What happens is that when you've got a goroutine waiting for events, it's possible to close it and create a new one, before other goroutines have a chance to respond. But since the previous descriptor was closed, the new inotify file descriptor will have the same number as the previous one, which will then get closed once the previous test's goroutines wake up and close their stuff.

So, yeah. From now on, Close waits for the readEvents goroutine to finish.

nathany:
Hah. "awesome" race conditions. So awesome that `-race` didn't catch it.

PieterD:
So if, during the execution of Add, some other goroutine closes the watcher and something else opens a file descriptor, it'll get messed up.

This changes a lot. Add and Remove will have to be properly synced with readEvents. Yeah, I don't think think you'll like the next pull request.

PieterD:
I wonder if -race could be adapted to catch this.

inotify.go
```diff
@@ -116,6 +136,21 @@ func (w *Watcher) Remove(name string) error {
 	return nil
 }
 
+// Called from Close(); remove all watches.
+func (w *Watcher) removeAll() int {
+	successful := 0
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	for name, watch := range w.watches {
+		success, _ := syscall.InotifyRmWatch(w.fd, watch.wd)
```
nathany: do we need `name = filepath.Clean(name)` here?

It was added to make ./path and path equivalent.

nathany: no, sorry, we don't use the name except for delete, so it's fine.

inotify.go
```diff
@@ -149,20 +183,31 @@ func (w *Watcher) readEvents() {
...
-			syscall.Close(w.fd)
-			close(w.Events)
-			close(w.Errors)
+			close(w.done)
 			return
 		}
 
 		if n < 0 {
-			w.Errors <- os.NewSyscallError("read", errno)
+			select {
```
nathany: these two selects make me curious.

we are sending an error or checking if the done channel is closed, whichever happens first?

inotify.go
PieterD: We don't need removeAll at all, actually. closing the fd will get rid of everything.

PieterD: In fact, we don't need watches at all.

inotify.go
nathany: so if the consumer of this library isn't ready to receive an Error or Event for some reason, we don't continue blocking if Close() is called.,

inotify_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestInotifyCloseSlightlyLaterWithWatch(t *testing.T) {
+	testDir := tempMkdir(t)
+	defer os.RemoveAll(testDir)
+
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("Failed to create watcher")
+	}
+	w.Add("testDir")
```
nathany: :bug: testDir without quotes?

inotify_test.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func TestInotifyCloseSlightlyLater(t *testing.T) {
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("Failed to create watcher")
+	}
+
+	// Wait until readEvents has reached syscall.Read, and Close.
+	<-time.After(50 * time.Millisecond)
```
nathany: Is there any advantage/difference of time.After vs. time.Sleep?

inotify.go
PieterD: Yeah. This solves the problem of when someone Closes the watcher and moves on, but there is still an error waiting to be sent. Without those selects, it would block forever. Which is, in fact, what was happening in a lot of cases.

PieterD: Er, anyway it still blocks on sending the error or event, unless we closed it. Only then will w.done return a value.

inotify_test.go
PieterD: Heh. OOPS

nathany: I can fix it, don't worry about it.

PieterD: Ok, thanks.

nathany:
For the changelog, #5 isn't actually solved yet, correct? (coming next)

PieterD:
I was under the impression that https://github.com/go-fsnotify/fsnotify/issues/5 was about goroutines blocking forever.

Perhaps I'm mistaken.

inotify.go
nathany: Don't we still need to remove things from our internal map? Otherwise a later Add will think it's already being watched.

inotify_test.go
PieterD: Not really.

nathany:
Squashed and merged. 

#5. Doesn't readEvent still block though? That little hack to wake it up on Close might work, but maybe not?

inotify.go
PieterD: We only use removeAll when we've already closed.

Ah, and we do need watches for the inotify handles. Sorry.

PieterD:
readEvents still has a blocking read on the file descriptor, but that's fine. So does kqueue and the Windows thing.

It should get cleaned up pretty much always, but not ABSOLUTELY always. That's the next one. So yeah, I suppose #5 might not be fixed until I push the epoll stuff.

nathany:
kqueue has a timeout right now so that it doesn't block indefinitely, but it does use a little bit of CPU to go around the loop every few 100ms.

Issue #59, opened 2015-01-27
File handle leak on OS X


neelance:
The following code quickly runs into a `panic: too many open files` on OS X, even though the documentation says that `Close` removes all watches.
```
package main

import "gopkg.in/fsnotify.v1"

func main() {
	for {
		println("loop")

		w, err := fsnotify.NewWatcher()
		if err != nil {
			panic(err)
		}

		go func() {
			for range w.Events {
			}
		}()
		go func() {
			for range w.Errors {
			}
		}()

		if err := w.Add("test.go"); err != nil {
			panic(err)
		}

		// uncomment this to make the leak go away
		// if err := w.Remove("test.go"); err != nil {
		// 	panic(err)
		// }

		if err := w.Close(); err != nil {
			panic(err)
		}
	}
}
```

nathany:
Thanks for reporting this.

nathany:
@PieterD I'm looking this on OS X and BSD, but it also fails on Linux:

using `import "github.com/go-fsnotify/fsnotify"`

```
panic: inotify_init: too many open files
```

PieterD:
This would cause a tremendous amount of problems. With my next pull, I've got w.Close waiting until the goroutine is done. That will fix this.

nathany:
Excellent.

**kqueue**

This is interesting, as Close just calls Remove for all watches. Another race condition? 
Close is ignoring the errors that Remove sends back, in this case: "bad file descriptor"

PieterD:
If Close just removes all watches but the goroutine blocks on something, the goroutines will start heaping up as you keep adding more goroutines, all blocking, all keeping open files.

They don't even have to block; if you create the next goroutine without waiting for the last one to finish, you could create a thousand before the first goroutine has a chance to clean up.

PieterD:
Confirmed that https://github.com/go-fsnotify/fsnotify/pull/66 fixes this for inotify. For OS X, I suggest making sure Close waits until the goroutine is stopped before returning.

nathany:
Thanks @PieterD. I've already merged a fix for OS X, but I have yet to tag a new release. Tomorrow once the poll stuff is in.

PR #65, opened 2015-02-07
kqueue: close kqueue after removing watches

nathany:
sending done would close w.kq before Remove had a chance to remove the watches with EV_DELETE, resulting in a file handle leak.

ref #59

also make Close() report last error returned by Remove.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -1109,6 +1109,21 @@ func TestConcurrentRemovalOfWatch(t *testing.T) {
 	<-removed2
 }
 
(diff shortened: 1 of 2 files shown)


nathany:
I'm not sure how I should test this. Based on #59, the second Add should reuse the same file descriptor because Close cleaned up. But I don't want to expose file descriptors to the API.

PieterD:
You can look at the internal field from Test.

nathany:
guess i can add a kqueue_test.go for that. ok

PR #66, opened 2015-02-07
Now use epoll so we can wake up readEvents

### LINKED ISSUE #5: inotify blocks
readEvent() currently blocks.

See https://code.google.com/p/go/issues/detail?id=5536

@howeyc attempted to change this in 3bfa915bde2415222d8dc048945f701353508160 but reverted it 7beb45183d2ab48bfc9616d9a019513a123a4983 due to reports of no new events https://github.com/howeyc/fsnotify/issues/52.

>  This was something I added to stop readEvent from blocking forever.
> Basically calling close would mean no new data came to the file descriptor
> and Read would block forever.
> 
> This was an attempt to use select with a timeout so it would return after
> the timeout stating whether there was new data.

### LINKED ISSUE #40: Cannot remove watch for deleted file
Once a file is deleted, you cannot clean up watches for it. For instance, with fsnotify:

``` go
	name = filepath.Clean(name)
	w.mu.Lock()
	defer w.mu.Unlock()
	watch, ok := w.watches[name]
	if !ok {
		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
	}
	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
	if success == -1 {
		return os.NewSyscallError("inotify_rm_watch", errno)
	}
	delete(w.watches, name)
	return nil
}
```

The file is gone, so even though we find the watch in `w.watches` we return a syscall error and don't delete the entry from `w.watches`.

It seems that maybe we should delete from `w.watches` regardless, or at least if `errno` is `EBADF`. If this sounds good, I can make a PR (and try to figure out the appropriate change for other systems (at least darwin) as well.

PieterD:
fixes #5.
fixes #40.
updates #59 (fixed for inotify).

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -194,27 +194,20 @@ func (w *Watcher) readEvents() {
 			return
 		}
 
-		// If EOF is received. This should really never happen.
-		if n == 0 {
-			select {
-			case w.Errors <- io.EOF:
-			case <-w.done:
-				return
-			}
-			continue
-		}
-
-		if n < 0 {
-			select {
-			case w.Errors <- errno:
-			case <-w.done:
-				return
-			}
-			continue
-		}
 		if n < syscall.SizeofInotifyEvent {
+			var err error
+			if n == 0 {
+				// If EOF is received. This should really never happen.
+				err = io.EOF
+			} else if n < 0 {
+				// If an error occured while reading.
+				err = errno
+			} else {
+				// Read was too short.
+				err = errors.New("notify: short read in readEvents()")
+			}
 			select {
-			case w.Errors <- errors.New("inotify: short read in readEvents()"):
+			case w.Errors <- err:
 			case <-w.done:
 				return
 			}
--- a/inotify.go
+++ b/inotify.go
@@ -28,7 +28,7 @@ type Watcher struct {
 	watches  map[string]*watch // Map of inotify watches (key: path)
 	paths    map[int]string    // Map of watched paths (key: watch descriptor)
 	done     chan struct{}     // Channel for sending a "quit message" to the reader goroutine
-	doneresp chan struct{}     // Channel to respond to Close
+	doneResp chan struct{}     // Channel to respond to Close
 }
 
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
@@ -52,7 +52,7 @@ func NewWatcher() (*Watcher, error) {
 		Events:   make(chan Event),
 		Errors:   make(chan error),
 		done:     make(chan struct{}),
-		doneresp: make(chan struct{}),
+		doneResp: make(chan struct{}),
 	}
 
 	go w.readEvents()
@@ -81,7 +81,7 @@ func (w *Watcher) Close() error {
 	w.poller.wake()
 
 	// Wait for goroutine to close
-	<-w.doneresp
+	<-w.doneResp
 
 	return nil
 }
@@ -155,7 +155,7 @@ func (w *Watcher) readEvents() {
 		ok    bool                                    // For poller.wait
 	)
 
-	defer close(w.doneresp)
+	defer close(w.doneResp)
 	defer close(w.Errors)
 	defer close(w.Events)
 	defer syscall.Close(w.fd)
--- a/inotify.go
+++ b/inotify.go
@@ -32,7 +32,13 @@ type Watcher struct {
 
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
 func NewWatcher() (*Watcher, error) {
-	poller, err := newFdPoller()
+	// Create inotify fd
+	fd, errno := syscall.InotifyInit()
+	if fd == -1 {
+		return nil, errno
+	}
+	// Create epoll
+	poller, err := newFdPoller(fd)
 	if err != nil {
 		return nil, err
 	}
--- a/inotify.go
+++ b/inotify.go
@@ -132,11 +132,20 @@ func (w *Watcher) Remove(name string) error {
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
+	// inotify_rm_watch will return EINVAL if the file has been deleted;
+	// the inotify will already have been removed.
+	// That means we can safely delete it from our watches, whatever inotify_rm_watch does.
+	delete(w.watches, name)
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
 	if success == -1 {
+		// TODO: Perhaps it's not helpful to return an error here in every case.
+		// the only two possible errors are:
+		// EBADF, which happens when w.fd is not a valid file descriptor of any kind.
+		// EINVAL, which is when fd is not an inotify descriptor or wd is not a valid watch descriptor.
+		// Watch descriptors are invalidated when they are removed explicitly or implicitly;
+		// explicitly by inotify_rm_watch, implicitly when the file they are watching is deleted.
 		return errno
 	}
-	delete(w.watches, name)
 	return nil
 }
 
--- a/inotify.go
+++ b/inotify.go
@@ -9,6 +9,7 @@ package fsnotify
 import (
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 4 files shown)


nathany:
I like how you separated the poller out into a separate file. But man, this makes it so much more complex.

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+// Create a new inotify poller.
+// This creates an inotify handler, and an epoll handler.
+func newFdPoller() (*fdPoller, error) {
+	var errno error
+	poller := new(fdPoller)
+
+	// Create inotify fd
+	poller.fd, errno = syscall.InotifyInit()
```
nathany: I think InotifyInit should remain in NewWatcher and pass the fd into newFdPoller.

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+	if poller.fd == -1 {
+		return nil, os.NewSyscallError("inotify_init", errno)
```
nathany: I found that wrapping with NewSyscallError breaks comparisons with the underlying error. Ended up removing them in 719fef588d53f4ccdf94803732cc34a6097ce1b7 but I missed this one.

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	// Create epoll fd
+	poller.epfd, errno = syscall.EpollCreate(1)
+	if poller.epfd == -1 {
+		syscall.Close(poller.fd)
+		return nil, os.NewSyscallError("epoll_create", errno)
+	}
+	// Create pipe; pipe[0] is the read end, pipe[1] the write end.
+	errno = syscall.Pipe(poller.pipe[:])
```
nathany: Should this be `err` rather than `errno` in this case?

inotify_poller.go
nathany: Likewise, NewWatcher should be responsible for closing the inotify fd if poller has an error.

Poller could in theory work for any fd passed in and only clean up its own resources.

inotify_poller.go
nathany: Maybe for EpollCtl too?

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	// Register pipe fd with epoll
+	event = syscall.EpollEvent{
+		Fd:     int32(poller.pipe[0]),
+		Events: syscall.EPOLLIN,
+	}
+	errno = syscall.EpollCtl(poller.epfd, syscall.EPOLL_CTL_ADD, poller.pipe[0], &event)
+	if errno != nil {
+		syscall.Close(poller.fd)
```
nathany: maybe cleanup could be put in a `defer func() {}` that checks if err is != nil? Might be cleaner, maybe not.

Maybe extracting a few functions, like to create and close the pipe, would help a bit?

inotify.go
```diff
@@ -188,10 +184,14 @@ func (w *Watcher) readEvents() {
 			return
 		}
 
-		// If EOF is received
+		// If EOF is received. This should really never happen.
```
nathany: :heart: :hear_no_evil:

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	return poller, nil
+}
+
+// Wait using epoll, then read from inotify.
+// Returns true if something is ready to be read,
+// false if there is not.
+func (poller *fdPoller) wait() (bool, error) {
+	events := make([]syscall.EpollEvent, 7)
```
nathany: why 7?

inotify_poller.go
```diff
@@ -0,0 +1,181 @@
...
+		if epollerr {
+			return true, nil
+		}
+		if epollhup || epollpipehup || epollpipein {
+			return false, nil
+		}
+		if epollin {
+			return true, nil
+		}
+		return false, errors.New("Epoll failed to generate any of the only six possibilities.")
```
nathany: oh. hm.

nathany:
Should there be inotify_poller_test.go?

inotify_poller.go
PieterD: There should be 3 possible events for each fd; hangup, error, and input. With two events, that's 6 maximum events if they all come in at once.

I don't know if epoll returns the number of events it wrote to the buffer, or the number of events TOTAL. But I'm catching both by making the buffer one larger than the maximum amount of events.

This should probably be a comment.

inotify.go
```diff
@@ -19,28 +20,39 @@ import (
...
-	done    chan struct{}     // Channel for sending a "quit message" to the reader goroutine
+	Events   chan Event
+	Errors   chan error
+	mu       sync.Mutex // Map access
+	fd       int
+	poller   *fdPoller
+	watches  map[string]*watch // Map of inotify watches (key: path)
+	paths    map[int]string    // Map of watched paths (key: watch descriptor)
+	done     chan struct{}     // Channel for sending a "quit message" to the reader goroutine
+	doneresp chan struct{}     // Channel to respond to Close
```
nathany: Maybe `doneResp` instead? Or `doneResposne`?

I was thinking of sending a signal one way over the done channel and then sending one back, but that won't work if the first signal is to close. Looking at the code more closely, I think we need to keep the two separate channels.

inotify.go
```diff
@@ -175,6 +167,20 @@ func (w *Watcher) readEvents() {
 			return
 		}
 
+		ok, errno = w.poller.wait()
+		if errno != nil {
+			select {
```
nathany: I'm starting to wonder if we should have a helper function to send an `error` down the channel or err/abort if done.

inotify_poller.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	errno = syscall.EpollCtl(poller.epfd, syscall.EPOLL_CTL_ADD, poller.pipe[0], &event)
+	if errno != nil {
+		return nil, errno
+	}
+
+	return poller, nil
+}
+
+// Wait using epoll, then read from inotify.
```
nathany: ~~then read form inotify~~

inotify_poller.go
```diff
@@ -0,0 +1,161 @@
...
+	n, errno := syscall.Read(poller.pipe[0], buf)
+	if n == -1 {
+		return errno
+	}
+	return nil
+}
+
+// Close all poller file descriptors, but not the one passed to it.
+func (poller *fdPoller) close() {
+	syscall.Close(poller.pipe[1])
```
nathany: if EpollCreate fails then pipe hasn't been used it. Is this going to break?

inotify_poller.go
```diff
@@ -0,0 +1,161 @@
...
+	buf := make([]byte, 1)
+	n, errno := syscall.Write(poller.pipe[1], buf)
+	if n == -1 {
+		return errno
+	}
+	return nil
+}
+
+func (poller *fdPoller) clearWake() error {
+	buf := make([]byte, 100)
```
nathany: why 100? :smirk:

inotify_poller_test.go
```diff
@@ -0,0 +1,221 @@
...
+	if errno != nil {
+		t.Fatalf("Failed to write to pipe: %v", errno)
+	}
+}
+
+func (tfd testFd) get(t *testing.T) {
+	buf := make([]byte, 10)
+	_, errno := syscall.Read(tfd[0], buf)
+	if errno != nil {
+		t.Fatalf("Failed to write from pipe: %v", errno)
```
nathany: read

inotify_poller.go
```diff
@@ -0,0 +1,161 @@
...
+
+type fdPoller struct {
+	fd   int    // File descriptor (as returned by the inotify_init() syscall)
+	epfd int    // Epoll file descriptor
+	pipe [2]int // Pipe for waking up
+}
+
+// Create a new inotify poller.
+// This creates an inotify handler, and an epoll handler.
+func newFdPoller(fd int) (_ *fdPoller, errno error) {
```
nathany: no need for the `_`

nathany: Oh. Mixing named and unamed. Mind you errno isn't really needed here for readability of the API. Could move it to var.

PieterD: But it is required to reference the return value in the deferred function two lines below.

PR #73, opened 2015-03-12
Inotify: fix watcher.watches and paths leaks

chamaken:
Thanks to @PieterD Watcher.Close() can stop goroutine cleanly.
But I think watches and paths map still leaks, and feel a bit strange
Remove() invalid filename returns differently between at first time and
after. I tried to fix those, would you revew the patches?

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -23,6 +23,7 @@ type Watcher struct {
 	Events   chan Event
 	Errors   chan error
 	mu       sync.Mutex // Map access
+	cv       *sync.Cond // sync removing on rm_watch with IN_IGNORE
 	fd       int
 	poller   *fdPoller
 	watches  map[string]*watch // Map of inotify watches (key: path)
@@ -54,6 +55,7 @@ func NewWatcher() (*Watcher, error) {
 		done:     make(chan struct{}),
 		doneResp: make(chan struct{}),
 	}
+	w.cv = sync.NewCond(&w.mu)
 
 	go w.readEvents()
 	return w, nil
@@ -132,20 +134,19 @@ func (w *Watcher) Remove(name string) error {
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
-	// inotify_rm_watch will return EINVAL if the file has been deleted;
-	// the inotify will already have been removed.
-	// That means we can safely delete it from our watches, whatever inotify_rm_watch does.
-	delete(w.watches, name)
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
 	if success == -1 {
-		// TODO: Perhaps it's not helpful to return an error here in every case.
-		// the only two possible errors are:
-		// EBADF, which happens when w.fd is not a valid file descriptor of any kind.
-		// EINVAL, which is when fd is not an inotify descriptor or wd is not a valid watch descriptor.
-		// Watch descriptors are invalidated when they are removed explicitly or implicitly;
-		// explicitly by inotify_rm_watch, implicitly when the file they are watching is deleted.
 		return errno
 	}
+
+	// not delete w.watches and w.paths here, but in ignoreLinux() called:
+	// InotifyRmWatch() above, readEvents() and ignoreLinux() chain
+	exists := true
+	for exists {
+		w.cv.Wait()
+		_, exists = w.watches[name]
+	}
+
 	return nil
 }
 
@@ -249,7 +250,7 @@ func (w *Watcher) readEvents() {
 			event := newEvent(name, mask)
 
 			// Send the events that are not ignored on the events channel
-			if !event.ignoreLinux(mask) {
+			if !event.ignoreLinux(w, raw.Wd, mask) {
 				select {
 				case w.Events <- event:
 				case <-w.done:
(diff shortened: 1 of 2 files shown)


PieterD:
Sure thing, I'll take a look at this tonight or tomorrow. Thanks.

PieterD:
Don't have much time (I'm on holiday!) but I've got a few minutes.

inotify.go
```diff
@@ -132,20 +134,19 @@ func (w *Watcher) Remove(name string) error {
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
-	// inotify_rm_watch will return EINVAL if the file has been deleted;
-	// the inotify will already have been removed.
-	// That means we can safely delete it from our watches, whatever inotify_rm_watch does.
```
PieterD: Not a big fan that all these comments were removed.

inotify.go
```diff
@@ -132,20 +134,19 @@ func (w *Watcher) Remove(name string) error {
...
-	delete(w.watches, name)
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
 	if success == -1 {
-		// TODO: Perhaps it's not helpful to return an error here in every case.
-		// the only two possible errors are:
-		// EBADF, which happens when w.fd is not a valid file descriptor of any kind.
-		// EINVAL, which is when fd is not an inotify descriptor or wd is not a valid watch descriptor.
-		// Watch descriptors are invalidated when they are removed explicitly or implicitly;
-		// explicitly by inotify_rm_watch, implicitly when the file they are watching is deleted.
 		return errno
```
PieterD: Or these

inotify.go
```diff
@@ -132,20 +134,19 @@ func (w *Watcher) Remove(name string) error {
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
-	// inotify_rm_watch will return EINVAL if the file has been deleted;
-	// the inotify will already have been removed.
-	// That means we can safely delete it from our watches, whatever inotify_rm_watch does.
-	delete(w.watches, name)
 	success, errno := syscall.InotifyRmWatch(w.fd, watch.wd)
```
PieterD: Oops! Was this a race condition? Looks like :)

PieterD:
Sorry, I don't have a ton of time :(

I'm a bit confused about what the patch is meant to fix. Could you elaborate?

chamaken:
Thank you for taking your time.
I will restore commens. How about changing first comment:
inotify_rm_watch will return EINVAL if the file has been deleted;
the inotify will already have been removed.
watches and pathes are deleted in ignoreLinux() implicitly and asynchronously
by calling inotify_rm_watch() below. e.g. readEvents() goroutine receives IN_IGNORE
so that EINVAL means that the wd is being rm_watch()ed or its file removed
by another thread and we have not received IN_IGNORE event.

inotify.go
chamaken: Would you tell me a story?

PieterD:
Thanks. There was some stuff in there I thought might save some time looking stuff up in the future :)

PieterD:
LGTM. I don't have a linux machine handy so I can't play around with it, but it looks good.

nathany:
Thanks @chamaken. I can take a look sometime over the next few days.

nathany:
Sorry, I still haven't checked this. :-(

tsuna:
This change fsnotify/fsnotify@e2532248b8b41873668e8e339833f9b4083d4a71 is making some of my unit tests fail non-deterministically.  After adding some tracing to the code, here's what I'm seeing (both on Linux 3.4.43 and 4.1.18):

When `Remove()` is called on a `Watcher`, it now `Wait()`s on a condition variable after calling `syscall.InotifyRmWatch()`.  The condition variable is notified from `ignoreLinux()`, at the point that we see an `Event` with the `IN_IGNORED` bit set.  This seems reasonable as the [man page for `inotify_rm_watch`](http://man7.org/linux/man-pages/man2/inotify_rm_watch.2.html) says:
> Removing a watch causes an `IN_IGNORED` event to be generated for this watch descriptor.

However I'm seeing that in two of my unit tests sometimes (more than 50% of the time) there is no `IN_IGNORED` event that follows the call to `syscall.InotifyRmWatch()`.  I verified that `syscall.InotifyRmWatch()` doesn't return any error.  There is just no event on the watch descriptor that was passed to `syscall.InotifyRmWatch()`.

Has anybody seen this before?

tsuna:
OK I think I got it... With this change it's no longer safe to consume Events from the same goroutine that calls `Remove()`.  In our case [our code](https://github.com/aristanetworks/tail/blob/master/watch/inotify_tracker.go#L144) basically looks this:

```go
func (shared *InotifyTracker) run() {
	// ...

	for {
		select {
		case winfo := <-shared.watch:
			shared.error <- shared.addWatch(winfo.fname)

		case fname := <-shared.remove:
			shared.removeWatch(fname) // <-------- no longer legit

		// ...
	}
}
```

Removing the watch leads to a deadlock: we're trying to call `Remove()` and `Remove()` is waiting to be notified that the `IN_IGNORED` event was received, but the `IN_IGNORED` event can't be sent because `fsnotify` is trying to write another event to the `Watch`.

nathany:
Thanks for investing this @tsuna. Would you mind opening a new issue so this doesn't get lost?

tsuna:
@nathany I'm happy to open an issue if you deem that this new code creates an undesirable side effect that needs to be fixed.  As you can see above, I've submitted a PR to workaround this quirk in the `tail` library.  But I agree that the new behavior of `fsnotify` as a result of this change is unfortunate because it will be more easily prone to deadlocks, depending on how the client code is written.

nathany:
Yes. I think we should take another look at this code change. Adding and removing watches from the same goroutine should not cause deadlocks.

tsuna:
I filed bug #123.  Thanks.

PR #76, opened 2015-04-27
Add Has* methods to simplify checking the state of an Event

mattrobenolt:
This helps expose a simpler way to query the state of an `Event` without resorting to the bitwise query on the raw opcode.

I originally named these as `Is*` instead of `Has*`, but went with `Has*` since each `Event` can hold multiple states. So `Has` seemed more appropriate.

I have signed the Google CLA.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -25,7 +25,7 @@ func ExampleNewWatcher() {
 			select {
 			case event := <-watcher.Events:
 				log.Println("event:", event)
-				if event.Op&fsnotify.Write == fsnotify.Write {
+				if event.HasWrite() {
 					log.Println("modified file:", event.Name)
 				}
 			case err := <-watcher.Errors:
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -30,25 +30,40 @@ const (
 	Chmod
 )
 
+// HasCreate returns true if Event has the Create opcode
+func (e Event) HasCreate() bool { return e.Op&Create == Create }
+
+// HasWrite returns true if Event has the Write opcode
+func (e Event) HasWrite() bool { return e.Op&Write == Write }
+
+// HasRemove returns true if Event has the Remove opcode
+func (e Event) HasRemove() bool { return e.Op&Remove == Remove }
+
+// HasRename returns true if Event has the Rename opcode
+func (e Event) HasRename() bool { return e.Op&Rename == Rename }
+
+// HasChmod returns true if Event has the Chmod opcode
+func (e Event) HasChmod() bool { return e.Op&Chmod == Chmod }
+
 // String returns a string representation of the event in the form
 // "file: REMOVE|WRITE|..."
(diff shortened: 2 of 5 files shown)


nathany:
Thanks for taking the time to contribute.

But I don't really want to add methods to the Event struct. My preference is to keep a minimal API surface with the bitfields exposed for checking, coalescing, etc. This is based on the design document mentioned in the README.

If you wants additional methods like this, it's easy enough to convert to a new type and provide them there.

An article on how to do this (and how bitfields work) would be a better approach IMO. Happy to take contributions to the fsnotify.org blog:
https://github.com/go-fsnotify/fsnotify.org

infogulch:
I feel like this would be much better if implemented on the `Op` type directly instead of on Event. I would say that it's similar to the [`os.FileMode`](https://golang.org/pkg/os/#FileMode) type, that's just a bit field int but has methods `IsDir` and `IsRegular` to extract info from it without using bitwise arithmetic directly.

As a bonus, if the Event struct is changed to embed Op anonymously these new methods on Op would be promoted to Event without cluttering its' explicit type (or documentation). I don't see any downside to this. Because the current name of the struct field is the same as the type name it should be 100% backwards compatible and references to Event.Op will still work as expected. Like so:
```
type Event struct {
	Name string // Relative path to the file or directory.
	Op          // File operation that triggered the event.
}
```

matryer:
I like this - but why Has? Why not Is?

IsCreate, IsWrite, etc. 

it also wouldn't stop switch code either.

omeid:
@matryer 
Perhaps because what _is_ Write can't be Remove, but what _has_ Write may have Remove too.

But overall, I agree with @nathany, I don't think any more methods should be added. 
This library should stay as minimal as possible and only solve the problem of cross-platform portability, if you want fancy helpers, you can just write a wrapper around this library and use it.

matryer:
Yeah. You'd almost do away with the bitmask consts in favour of the methods. 

But, it's pretty nice as it is too.

omeid:
IMHO, the problem with adding these methods is that you either have to unexport the `Event.Op` (breaking change, methods are more restricted than bitmask) or have overlapping API. 

Now overlapping APIs doesn't so bad until you consider that there are 8 backends, so where you need 8 tests, with this new methods you need at least 16, where a feature could have been inconsistent or broken in 8 places, now it can be at least 16.

This is unnecessary complexity that could be shifted into it's own package.

matryer:
It's more an academic discussion about package API design really. I agree that there's no need to change it now.

nathany:
The use of bitmasks and creation of a wrapper could be documented on the blog. Would anyone like to write it up? https://github.com/go-fsnotify/fsnotify.org

nathany:
I'm closing this in favour of documentation or a blog post. See #107. Pull requests welcome.

PR #86, opened 2015-08-15
Add string method to event.Op

techjanitor:


omeid:
This has a fair bit of overlapping with `Event.String()`, you should simplify `Event.String()` if you want to add `Op.String()`.

techjanitor:
I can, but, Event.String() doesn't make much sense to me period. Name is already a string, why not just let people format their own output .. ? I don't know what the intent of it is. It doesn't look like it's used internally.

matryer:
It's great for debugging. Shows you the op and path.

omeid:
As explained by @matryer , the `fmt` package which is used with the `log` and other logging packages respects the `Stringer` interface. Having `Event.String() string` means that you can simply do `log.Print(event)` and you would get something nice according to the `Event.String() string` output.

PR #88, opened 2015-08-17
Add Op.String() and fix Event.String() to use Op.String()

techjanitor:
Event.String() seemed to, at one point in time, return a list of events? It looks like it there can only be one Op in an event so..

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -47,9 +47,7 @@ func (o Op) String() string {
 	return ""
 }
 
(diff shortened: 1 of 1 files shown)


nathany:
There can be multiple Ops in an event. That's why it is a bitmask.

nathany:
@techjanitor for your info, #165 has been merged, resolving the issue you set out to with this pull request.

Issue #22, opened 2014-06-29
Better handling of file-close events on Linux


nathany:
Requested by @spf13 and @extemporalgenome.

> @spf13: on linux at least, all you need to do the proper tracking of vim/any-other-kind-of-editor "done writing" tracking is watch for IN_CLOSE_WRITE and IN_MOVE_TO on the directory
> 
> @extemporalgenome if it's a lazy editor, that just updates in place, IN_CLOSE_WRITE catches it; if it's a "write to temporary and atomically rename" editor, IN_MOVE_TO catches it

The trouble is in doing this in a cross-platform way. And some users may still want IN_MODIFY (eg. to show the progress of a file writing).

stumpyfr:
Any new on that?

nathany:
Nothing yet.

edrex:
I'm guessing @extemporalgenome's original question was "How can an fsnotify user listen for changes to a single file in a way that works with all [the various text editors' save strategies](https://github.com/guard/guard/wiki/Analysis-of-inotify-events-for-different-editors)?".

Many editors are do a little dance around saving which confounds the strategy of listening to the file for write events. For example, `vim`:

```
2015/03/25 23:18:42 event: "./foo.txt": RENAME
2015/03/25 23:18:42 event: "./foo.txt~": CREATE
2015/03/25 23:18:42 event: "./foo.txt": CREATE
2015/03/25 23:18:42 event: "./foo.txt": WRITE
2015/03/25 23:18:42 modified file: ./foo.txt
```

It seems like the most compatible strategy is to listen to the directory and only pay attention to events matching the file name, but it's not clear which event types to listen to. Just `fsevents.Write` is enough for vim. Any recommendation?

nathany:
This is closely related to #17.

rjeczalik:
Hey @edrex, @nathany! If'd want to use `IN_MOVE_{FROM,TO}` events for inotify, then notify supports them out of the box - see `LinuxMove` example for the [Watch](https://godoc.org/github.com/rjeczalik/notify#Watch) method, in particular:

```go
err := notify.Watch("/dir/to/watch", channel, notify.InMovedFrom, notify.InMovedTo);
```

I'm not suggesting you should pick it over fsnotify, I'm suggesting you could at least test your use-case, whether you really need platform specific events for that.

edrex:
@nathany thanks, #17 is just what I was looking for.

[...]

PR #91, opened 2015-09-01
Added new event type CLOSED

manatok:
Added a new event type 'CLOSED' to signify that a file write has closed. Listening for the IN_CLOSE_WRITE

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -9,7 +9,7 @@ package fsnotify_test
 import (
 	"log"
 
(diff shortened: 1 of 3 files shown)


nathany:
Hi David,
Thanks for the pull request. Is this meant to address #22?

The thing that I'm not sure about is operating systems other than Linux. With this change, there is a Closed event that doesn't mean anything outside of Linux. Hm.

nathany:
I hope you don't mind working off a fork until we come up with a better solution.

PR #98, opened 2015-10-14
Don't try to watch named pipes

evanphx:
You can't watch a named pipe, same as a socket. The call to `Open` will just hang forever.


nathany:
Thanks Evan. 

It looks like there is an unrelated problem in CircleCI and Travis. I'll look into it.

nathany:
Merged to master. Thanks.

evanphx:
@nathany Any chance this could be cherry-picked into v1?

nathany:
Yup. v1.2.1 released with this change.

https://gopkg.in/fsnotify.v1

evanphx:
Thank you!

PR #100, opened 2015-11-02
switch to epoll_create1

suihkulokki:
Epoll_create was replaced with epoll_create1 in linux 2.6.27. The older
syscall isn't available on new architectures (such as arm64). Without this,
error: docker[1582]: syscall 1042 on kernel logs while docker fails to start.

CLA has been signed and code tested on linux (this code doesn't run anywhere else)

### DIFF
--- a/inotify_poller.go
+++ b/inotify_poller.go
@@ -39,7 +39,7 @@ func newFdPoller(fd int) (*fdPoller, error) {
 	poller.fd = fd
 
 	// Create epoll fd
-	poller.epfd, errno = syscall.EpollCreate(1)
+	poller.epfd, errno = syscall.EpollCreate1(0)
 	if poller.epfd == -1 {
 		return nil, errno
 	}


nathany:
Thanks.

Go itself only [requires Linux 2.6.23 or later with glib](https://golang.org/doc/install), so I should add a note to the README about 2.6.27.

omeid:
What is the stats on kernel versions usage? If this will upset people, I think one can make it backward compatible with direct `syscall.RawSyscall`, First trying the old SYS_EPOLL_CREATE/`1042` and then SYS_EPOLL_CREATE1/`291`?

nathany:
I'm not sure what the stats are. That's a good idea to look into it. Preferably we won't need a kludge to support the older kernels, but mentioning it in the README and release notes is important.

suihkulokki:
2.6.27 was released already in 2008 -  2.6.32 is generally the last relevant kernel to support (it's the kernel RHEL6 from 2010 has).

suihkulokki:
Hi, is there something needed from my side for this patch?

nathany:
No, sorry for the delay. 

I'll merge some things and do a release later today.

nathany:
Merged. Thanks Riku.

I'l do a new point release after looking at a few other pull requests.

PR #101, opened 2015-11-17
Avoid infinite loops from symlinks cycles

illicitonion:
The semenatics of using kqueue are already slightly different than
inotify, specifically that inotify will give you events for the symlink
itself whereas kqueue will only give events for the target of the
symlink. This preserves that behaviour, that symlinks don't get notified
on, but ensures that circular symlinks don't recurse forever.

This change pushes the resolved filename up through the stack so that
the fileExists map can be properly populated; not doing so would mean
that deleting a cyclical symlink and then replacing it with a file of
the same name would not notify for that file's creation (or subsequent
events) because we would believe that the file already existed.

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -94,7 +94,8 @@ func (w *Watcher) Add(name string) error {
 	w.mu.Lock()
 	w.externalWatches[name] = true
 	w.mu.Unlock()
-	return w.addWatch(name, noteAllEvents)
+	_, err := w.addWatch(name, noteAllEvents)
+	return err
 }
 
 // Remove stops watching the the named file or directory (non-recursively).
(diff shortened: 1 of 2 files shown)


illicitonion:
Fixes https://github.com/go-fsnotify/fsnotify/issues/96

illicitonion:
I'm not sure why that second commit made the linux CI fail; none of the tests failed, and running the tests on a Linux machine I have they all pass and exit 0 :S

nathany:
Thanks for the PR. 

Regarding the Linux test failure, I've triggered a rebuild, so we'll see if Travis is just being flaky.

kqueue.go
```diff
@@ -153,7 +154,8 @@ var keventWaitTime = durationToTimespec(100 * time.Millisecond)
 
 // addWatch adds name to the watched file set.
 // The flags are interpreted as described in kevent(2).
-func (w *Watcher) addWatch(name string, flags uint32) error {
+// Returns the real path to the file which was added, if any, which may be different from the one passed in the case of symlinks.
```
nathany: I don't see where this new return value is being used.

kqueue.go
```diff
@@ -195,18 +197,26 @@ func (w *Watcher) addWatch(name string, flags uint32) error {
...
 			}
 
 			fi, err = os.Lstat(name)
 			if err != nil {
-				return nil
+				return "", nil
+			}
+
+			w.mu.Lock()
+			_, alreadyWatching = w.watches[name]
```
nathany: :+1: 

this check could probably even come before the Lstat.

kqueue.go
illicitonion: Lines 378 and 384

kqueue.go
illicitonion: Done

Issue #99, opened 2015-10-21
Date race


djui:
While race testing some of my code I got the following. Seems like watcher are being close although they have been already removed...

```
WARNING: DATA RACE
Write by goroutine 101:
  runtime.mapdelete()
      /usr/local/Cellar/go/1.5.1/libexec/src/runtime/hashmap.go:511 +0x0
  ./vendor/gopkg.in/fsnotify%2ev1.(*Watcher).Remove()
      ./vendor/gopkg.in/fsnotify.v1/kqueue.go:119 +0x4ed
  ./vendor/gopkg.in/fsnotify%2ev1.(*Watcher).readEvents()
      ./vendor/gopkg.in/fsnotify.v1/kqueue.go:291 +0x878

Previous read by goroutine 189:
  runtime.mapiternext()
      /usr/local/Cellar/go/1.5.1/libexec/src/runtime/hashmap.go:621 +0x0
  ./vendor/gopkg.in/fsnotify%2ev1.(*Watcher).Close()
      ./vendor/gopkg.in/fsnotify.v1/kqueue.go:80 +0x233
  ./vendor/gopkg.in/fsnotify%2ev1.(*Watcher).Close-fm()
      ./foo/foo.go:157 +0x3b
  ..LogCloseError()
      ./log_close_error.go:9 +0x48
  ./foo.(*fooReader).monitor.func1()
      ./foo/foo.go:178 +0x512

Goroutine 101 (running) created at:
  ./vendor/gopkg.in/fsnotify%2ev1.NewWatcher()
      ./vendor/gopkg.in/fsnotify.v1/kqueue.go:61 +0x443
  ./foo.(*fooReader).monitor()
      ./foo/foo.go:146 +0x58
  ./foo.New()
      ./foo/foo.go:48 +0x31b
  ./foo_test.TestClosing()
      ./foo/foo_test.go:118 +0x3e2
  testing.tRunner()
      /usr/local/Cellar/go/1.5.1/libexec/src/testing/testing.go:456 +0xdc

Goroutine 189 (running) created at:
  ./foo.(*fooReader).monitor()
      ./foo/foo.go:181 +0x228
  ./foo.New()
      ./foo/foo.go:48 +0x31b
  ./foo_test.TestClosing()
      ./foo/foo_test.go:118 +0x3e2
  testing.tRunner()
      /usr/local/Cellar/go/1.5.1/libexec/src/testing/testing.go:456 +0xdc
==================
[...]

PR #102, opened 2015-11-25
Add test for concurrent file removes and watcher.Close calls. See issue #99

ppknap:
#99


nathany:
Should add `script: go test -v --race ./...` to use the race detector on Travis.

nathany:
I'm closing this as it's included in #105, which is on the race branch. Thanks.

Issue #17, opened 2014-06-29
Atomic saves in text editors cause havoc, unable to watch individual files


nathany:
Reported by @robfig at https://github.com/howeyc/fsnotify/issues/91. See also https://github.com/nathany/looper/issues/14.

When editors use _atomic saves_ they Create a new file and then do a Rename rather than just a straight Write (like when touching a file).

Related issues:
- @mstum wanted to watch a single configuration file (https://github.com/howeyc/fsnotify/issues/94), and @iyf also had trouble watching a single file (https://github.com/howeyc/fsnotify/issues/37).
- If you Add a directory and then Remove a single file within the directory, the watcher will still report a Create due to atomic saves. https://code.google.com/p/go/issues/detail?id=7788

Dealing with text editors may be out of scope of fsnotify itself, but it could at least provide recommendations in the Wiki. 

@robfig also reported some other oddities, though those may be specific to kqueue on OS X.

nathany:
TextMate2 (OS X) use the exchangedata system call to perform the saves (not sure what that is). But Hugo users are having problems detecting saves: https://github.com/spf13/hugo/issues/1053#issuecomment-116778527

arp242:
Closing this as a duplicate of #372, since it's essentially the same problem. That issue is much newer, but also more comprehensive.

PR #111, opened 2016-01-10
Fix logic for CREATE after REMOVE (for TextMate etc.)

bep:
`os.Lstat` does not return `error` when a file/directory exists, so the code path  to check for newly created files after a REMOVE on BSD was never executed.

Also, if the watched entry is a file, it has been removed from the watcher at this point, so do a new-file-check for them as well.

This commit fixes issues on TextMate on OS X, and probably others (`exchangedata` is used in `NSDocument`), that uses `exchangedata` to do atomic saves.

See #17

### DIFF
--- a/integration_darwin_test.go
+++ b/integration_darwin_test.go
@@ -1,0 +1,146 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package fsnotify
+
+import (
+	"os"
+	"path/filepath"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// testExchangedataForWatcher tests the watcher with the exchangedata operation on OS X.
+//
+// This is widely used for atomic saves on OS X, e.g. TextMate and in Apple's NSDocument.
+//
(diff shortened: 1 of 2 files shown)


bep:
Note: I will add a test case for this shortly.

bep:
I have added a test case + signed the Google Individual Contributor License Agreement.

bep:
There are some obvious issues with one part of this commit (checking for new files when the watched entry is a file and not a directory) -- I will check if I can fix it, or revert that part.

bep:
OK - now it works fine.

I have tested this OK on both of the trouble-arease in Hugo (config reloading (file based) and content watching (dir based)).

renatovargas:
Awesome, great work!!

bep:
On an added note:

I have fiddled a little with this on the Linux side of it as well, today, and on `Linux` the only way to track changes from Vim (and similar) is to watch an entire folder, then pick the file and event type you are interested in (it will eventually be a CREATE). This is how inotify works, and I don't think fsnotify should "write some custom code" to get around it.

But with this PR, this means that BSD != Linux in the single file case, and that is maybe worth a thought. Should it? 

If they should work the same, this patch (excluding test) will be reduced to the `err == nil` conditional..

anthonyfok:
+1  You're awesome, bep!  :-)

nathany:
Thanks @bep

PR #113, opened 2016-01-15
Add stringify functionality for better pretty printing

purpleidea:


purpleidea:
w00t! Build is now green. Enjoy!

omeid:
See #86

purpleidea:
@omeid Looks similar, but of course my patch is more correct. I don't see what the problem is, this is a useful thing.

purpleidea:
To add some context, if you print out the Op, you just get a useless number instead of a useful string:

https://github.com/purpleidea/mgmt/blob/master/file.go#L158

omeid:
Nothing wrong, but if you want to go this way, you need to refactor [Event.String()](https://github.com/go-fsnotify/fsnotify/blob/master/fsnotify.go#L35-L53) to use `op.String()` as well.

purpleidea:
@omeid Oh, I'm sorry. I misunderstood what the other thread was saying. I'm happy to refactor it, and I will do it in a moment. Do you wish to require that the strings stay uppercase as to avoid changing that string interface, or can we switch them to the same case as in my Op. (My preference.)

purpleidea:
@omeid Patch updated and passes tests on my machine. The failure seems to be some bug in the build system or an issue that I don't understand. The error is:

```
./fsnotify.go:44: o.String undefined (type Op has no field or method String)
```

But I call the go generate function, and it WFM.

purpleidea:
Ping?

fsnotify_test.go
```diff
@@ -0,0 +1,51 @@
...
+// Package fsnotify provides a platform-independent interface for file system notifications.
+package fsnotify
+
+import (
+	"testing"
+)
+
+func TestOpStringify(t *testing.T) {
+	if Create.String() != "Create" {
+		t.Fatalf("Stringify failed for: Create Op.")
```
nathany: You can just use t.Error() here, reporting on any errors rather than failing immediately.

fsnotify_test.go
```diff
@@ -0,0 +1,51 @@
...
+		t.Fatalf("Stringify failed for: Rename Op.")
+	}
+	if Chmod.String() != "Chmod" {
+		t.Fatalf("Stringify failed for: Chmod Op.")
+	}
+}
+
+func TestEventStringify(t *testing.T) {
+	e0 := &Event{"foo0", 0}
+	// TODO: while the foo0 output is consistent with the previous
```
nathany: Feel free to come up with more elegant output.

.gitignore
```diff
@@ -4,3 +4,4 @@
 
 .vagrant
 *.sublime-project
+*_stringer.go
```
nathany: please check in op_stringer.go instead of excluding the file.

please don't use go generate on Travis either.

nathany:
Thanks @purpleidea. I left some comments.

purpleidea:
Hi @nathany I "tunnel vision" never thought to use Errorf instead of Fatalf. Good suggestion, fixed!

I also made the other changes.

As an aside, I prefer to generate the Stringer stuff instead of committing it, so I was wondering why you would like to do it this way? In any case, it's in my patch the way you preferred.

Cheers!

fsnotify_test.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build !plan9,!solaris,!windows
```
nathany: why not on Windows?

fsnotify_test.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	e1 := &Event{"foo1", Chmod}
+	if s1, s2 := `"foo1": Chmod`, e1.String(); s1 != s2 {
+		t.Errorf("Stringify expected: `%s`; got: `%s`.", s1, s2)
+	}
+	e2 := &Event{"foo2", Write + Chmod}
+	if s1, s2 := `"foo2": Write|Chmod`, e2.String(); s1 != s2 {
+		t.Errorf("Stringify expected: `%s`; got: `%s`.", s1, s2)
+	}
+	e3 := &Event{"foo3", Create + Write + Remove + Rename + Chmod}
```
nathany: Even though this works, I think using bitwise & makes more sense because they are bits. :smile:

purpleidea:
@nathany 
Doesn't work for me:

```
--- FAIL: TestEventStringify (0.00s)
	fsnotify_test.go:43: Stringify expected: `"foo2": Write|Chmod`; got: `"foo2": nil`.
	fsnotify_test.go:47: Stringify expected: `"foo3": Create|Write|Remove|Rename|Chmod`; got: `"foo3": nil`.
FAIL
exit status 1
FAIL	_/home/james/code/src/fsnotify	13.449s
```

nathany:
Thanks James. As to your question, if someone does a `go get github.com/go-fsnotify/fsnotify` the code should compile and build without needing to go generate anything. 

This applies even more so to a library. For example, if you `go get github.com/spf13/hugo` which depends on fsnotify, it would fail to build until you run go generate on the fsnotify dependency. If multiple libraries did that it would be a total nightmare for users.

It's also what the Go team recommends, and I trust their recommendations, because they have a lot more experience than me. https://blog.golang.org/generate

nathany:
Is that test failure on Windows?

purpleidea:
@nathany Good point RE: go generate. In this case if there's no mechanism to run this automatically, then I consider this an issue with `go get`.

As for the test failure, it is currently passing. I don't have a Windows machine.
At this point if there's something else you want to modify, feel free to just add a patch on top of mine if it would make it easier for you.

Cheers

nathany:
I'm too sleepy to be coding on this right now (bugs are inevitable), but I'll try to take a look in the next few days. Thanks for the contrib.

purpleidea:
@nathany ping :(

nathany:
pong :)

sorry.

nathany:
This PR still needs a little more work before it's merged. I can do the work, but I haven't had a chance to yet.

purpleidea:
@nathany What's missing?

nathany:
see the comments i made during code review.

purpleidea:
@nathany

> see the comments i made during code review.

I don't know what's missing... 

> This PR still needs a little more work before it's merged. I can do the work, but I haven't had a chance to yet.

This was just a "get my feet wet" patch for this project. Feel free to do with it what you want or not. Cheers!

purpleidea:
Patch is here: https://github.com/purpleidea/fsnotify/tree/feat/stringify if you want it. I'm going to close this for now.

nathany:
ok.

Issue #24, opened 2014-07-06
kqueue improvements


nathany:
What have I learned from experimenting with kqueue directly (primarily on OS X)? #23 

- [x] fsnotify doesn't watch for Extend, Link or Revoke events (tracked in #519 now).

- [x] fsnotify has a 100ms timeout on reads, but instead we may be able to just block until kqueue has events available (nil timespec). this could save a little CPU time when idle, see: https://github.com/howeyc/fsnotify/issues/58
  - closing the kq (on OS X) results in the read failing with _Kevent: interrupted system call_, so blocking in a goroutine shouldn't delay shutdown.

- [x] fsnotify could add multiple file descriptors at once (such as all files in a directory), which may be a little more efficient then adding each one individually.

- [x] investigated `Udata` which C programs sometimes use to point to the watched filename, but it's probably not worthwhile from Go (and NetBSD defines Udata differently than the other BSDs).

Other improvements are specific to fsnotify, such as less mutexes #13 and bug fixes #14.

Platin21:
I always wondered how it deals with atomic writes as i cannot get it to fire again when such a write happens.

arp242:
> fsnotify could add multiple file descriptors at once (such as all files in a directory), which may be a little more efficient then adding each one individually.

`Add()` + `Remove()` takes 10078ns on my slow VM inside my slow laptop, and I bet most of that is in file operations rather than the syscall:

    BenchmarkAddRemove-2      128599             10078 ns/op             848 B/op         13 allocs/op

Not really worth optimizing IMO.

Issue #89, opened 2015-08-26
watcher.Close() is hanging forever on OSX


ryanwalls:
I'm using the standard pattern presented in the samples file.  Running the watcher from within a unit test and the unit test never exits when watcher.Close() is called.

Let me know what details would be useful.

nathany:
If you can share some sample code to reproduce the issue, that would be helpful..

omeid:
The example doesn't check if `watcher.Events` so after calling `watcher.Close()` the outer select would block forever, assuming there are no errors.

pquerna:
I was having similar issues on OSX, and PR #124 fixed them for me.

nhooyr:
I believe this is fixed in #230 @nathany

nhooyr:
Never mind, seems unrelated if the OP is following the example.

PR #124, opened 2016-03-04
kqueue: wait without timeout

fjl:
As noted in #24, reading from the kqueue won't unblock when
it is closed on all platforms. This commit solves the issue by
additionally watching on a pipe which is closed when Watcher
is closed. This makes the read timeout unnecessary.

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -13,7 +13,6 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
-	"time"
 
 	"golang.org/x/sys/unix"
 )
@@ -22,9 +21,9 @@ import (
 type Watcher struct {
 	Events chan Event
 	Errors chan error
-	done   chan bool // Channel for sending a "quit message" to the reader goroutine
 
-	kq int // File descriptor (as returned by the kqueue() syscall).
+	kq        int    // File descriptor (as returned by the kqueue() syscall).
+	closepipe [2]int // Pipe used for closing.
 
 	mu              sync.Mutex        // Protects access to watcher data
 	watches         map[string]int    // Map of watched file descriptors (key: path).
@@ -42,13 +41,14 @@ type pathInfo struct {
 
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
 func NewWatcher() (*Watcher, error) {
-	kq, err := kqueue()
+	kq, closepipe, err := kqueue()
 	if err != nil {
 		return nil, err
 	}
 
 	w := &Watcher{
 		kq:              kq,
+		closepipe:       closepipe,
 		watches:         make(map[string]int),
 		dirFlags:        make(map[string]uint32),
 		paths:           make(map[int]pathInfo),
@@ -56,7 +56,6 @@ func NewWatcher() (*Watcher, error) {
 		externalWatches: make(map[string]bool),
 		Events:          make(chan Event),
 		Errors:          make(chan error),
-		done:            make(chan bool),
 	}
 
 	go w.readEvents()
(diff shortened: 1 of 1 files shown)


fjl:
I cannot reproduce the failure on Travis and it seems unrelated to the PR.

fjl:
The tests pass on FreeBSD, too.

nathany:
I'll re-run Travis to see if it's being flaky.

fjl:
Looks like it worked this time

fjl:
I found a minor issue with how the pipe is set up and fixed it. Travis still passing.

fjl:
Any ETA on getting this merged?

nathany:
Sorry, still haven't reviewed it yet.

pquerna:
I was also seeing Close() hangs as described in #89 -- at $job we are now carrying this patch on our vendor'ed copy of fsnotify, would love to see it merged.

nathany:
I'm not that familiar with `syscall.Pipe`. Hoping someone will review this.

fjl:
rebased

fjl:
Is there anything I can do to get this merged?

nathany:
There is currently a data race issue with kqueue on master, which should be remedied first.
https://github.com/fsnotify/fsnotify/pull/162#issuecomment-251575890

As to this PR, we need some familiar with epoll to do the code review.

arp242:
A lot of things changed since this PR was opened and it doesn't rebase cleanly with quite a few conflicts. It's not entirely clear to me if this problem has been solved since 2014 or if it still exists.

Either way, I think it would be good to have a test case; I ran `go test -race -count=100` on FreeBSD (13) and that runs without problems, but maybe it only happens on macOS?

nathany:
@arp242 GitHub Actions will run on macOS, if there is a useful test case, but not sure about running such a high count.

On macOS 12.4:

> ❯ go test -race -count=100
> panic: test timed out after 10m0s

arp242:
Oh, you'll have to add `-timeout=20m` or some such; the timeout applies to the entire test run, not a single test, so if that takes too long it will time out.

> GitHub Actions will run on macOS, if there is a useful test case, but not sure about running such a high count.

Yeah, just a quick low-effort way to see if intermittent problems show up; if they do, a dedicated (non-slow) test case can be added.

nathany:
This is the result running another time on macOS 12.5:

```
❯ go test -race -count=100 -timeout=20m
--- FAIL: TestFsnotifyDeleteWatchedDir (0.52s)
    integration_test.go:460: error received: fdopendir /var/folders/82/9wkthct972xb6h3nwsx2rfnh0000gn/T/fsnotify1522497754: no such file or directory
    integration_test.go:471: event received: "/var/folders/82/9wkthct972xb6h3nwsx2rfnh0000gn/T/fsnotify1522497754/TestFsnotifyEventsExisting.testfile": REMOVE
    integration_test.go:471: event received: "/var/folders/82/9wkthct972xb6h3nwsx2rfnh0000gn/T/fsnotify1522497754": REMOVE
FAIL
exit status 1
FAIL	github.com/fsnotify/fsnotify	886.645s
```

arp242:
Coolio, I can't reproduce that on FreeBSD (just tried again to be sure), but it confirms that the problem still exists, and that it's specific to macOS and not all of kqueue (*probably*).

So need to rebase and write a decent test case which doesn't take 15 minutes, probably something [similar to this](https://github.com/fsnotify/fsnotify/pull/465/files#diff-ed1088fda4ca38792b99302b355705eb525017fca294e83858422bb430d8cc59R1247).

arp242:
#262 seems to do the same btw, might want to go with that one as it seems a bit simpler.

arp242:
New version of this PR: #480

I couldn't update this PR for some reason; I guess the branch/fork disappeared after all these years? Anyway, it's essentially the same patch, but rebased on the latest main and with some minor cleanups.

Issue #130, opened 2016-03-28
Does not work in powerpc64 or arm64


TheTincho:
Hi,

I package prometheus for Debian. And I am getting build errors in ppc64el, which seem to originate in fsnotify. If I try to run fsnotify's tests there I get all kind of failures, and a hang:

```
github.com/go-fsnotify/fsnotify
   dh_auto_test -O--buildsystem=golang
	go test -v github.com/go-fsnotify/fsnotify
=== RUN   TestPollerWithBadFd
--- PASS: TestPollerWithBadFd (0.00s)
=== RUN   TestPollerWithData
--- FAIL: TestPollerWithData (0.00s)
	inotify_poller_test.go:85: expected poller to return true
=== RUN   TestPollerWithWakeup
--- PASS: TestPollerWithWakeup (0.00s)
=== RUN   TestPollerWithClose
--- FAIL: TestPollerWithClose (0.00s)
	inotify_poller_test.go:119: expected poller to return true
=== RUN   TestPollerWithWakeupAndData
--- FAIL: TestPollerWithWakeupAndData (0.00s)
	inotify_poller_test.go:140: expected poller to return true
=== RUN   TestPollerConcurrent
--- FAIL: TestPollerConcurrent (0.05s)
	inotify_poller_test.go:197: expected true
=== RUN   TestInotifyCloseRightAway
--- PASS: TestInotifyCloseRightAway (0.05s)
=== RUN   TestInotifyCloseSlightlyLater
--- PASS: TestInotifyCloseSlightlyLater (0.10s)
=== RUN   TestInotifyCloseSlightlyLaterWithWatch
--- PASS: TestInotifyCloseSlightlyLaterWithWatch (0.10s)
=== RUN   TestInotifyCloseAfterRead
--- PASS: TestInotifyCloseAfterRead (0.10s)
=== RUN   TestInotifyCloseCreate
--- FAIL: TestInotifyCloseCreate (0.05s)
	inotify_test.go:136: Took too long to wait for event
=== RUN   TestInotifyStress
--- FAIL: TestInotifyStress (5.00s)
	inotify_test.go:238: Expected at least 50 creates, got 0
=== RUN   TestInotifyRemoveTwice
--- PASS: TestInotifyRemoveTwice (0.00s)
=== RUN   TestInotifyInnerMapLength
  
signal: terminated
FAIL	github.com/go-fsnotify/fsnotify	348.362s
```

TheTincho:
[...]

Issue #112, opened 2016-01-13
Add support for Linux/arm64 (aarch64)


anthonyfok:
After initial diagnosis of spf13/hugo#1772, I discovered that fsnotify is using some syscalls that are unavailable on the Linux for ARM64 platform.  Here is the result of `go test -v ./...` run:

```
=== RUN   TestPollerWithBadFd
--- PASS: TestPollerWithBadFd (0.00s)
=== RUN   TestPollerWithData
--- FAIL: TestPollerWithData (0.13s)
	inotify_poller_test.go:82: poller failed: function not implemented
=== RUN   TestPollerWithWakeup
--- FAIL: TestPollerWithWakeup (0.13s)
	inotify_poller_test.go:101: poller failed: function not implemented
=== RUN   TestPollerWithClose
--- FAIL: TestPollerWithClose (0.13s)
	inotify_poller_test.go:116: poller failed: function not implemented
=== RUN   TestPollerWithWakeupAndData
--- FAIL: TestPollerWithWakeupAndData (0.13s)
	inotify_poller_test.go:137: poller failed: function not implemented
=== RUN   TestPollerConcurrent
--- FAIL: TestPollerConcurrent (0.13s)
	inotify_poller_test.go:180: poller failed: function not implemented
	inotify_poller_test.go:197: expected true
=== RUN   TestInotifyCloseRightAway
--- PASS: TestInotifyCloseRightAway (0.05s)
=== RUN   TestInotifyCloseSlightlyLater
--- PASS: TestInotifyCloseSlightlyLater (0.18s)
=== RUN   TestInotifyCloseSlightlyLaterWithWatch
--- PASS: TestInotifyCloseSlightlyLaterWithWatch (0.18s)
=== RUN   TestInotifyCloseAfterRead
--- PASS: TestInotifyCloseAfterRead (0.18s)
=== RUN   TestInotifyCloseCreate
--- FAIL: TestInotifyCloseCreate (0.13s)
	inotify_test.go:136: Took too long to wait for event
=== RUN   TestInotifyStress
--- FAIL: TestInotifyStress (0.13s)
	inotify_test.go:242: Got an error from watcher: function not implemented
=== RUN   TestInotifyRemoveTwice
--- PASS: TestInotifyRemoveTwice (0.00s)
=== RUN   TestInotifyInnerMapLength
fatal error: all goroutines are asleep - deadlock!

goroutine 1 [chan receive]:
[...]

PR #135, opened 2016-04-13
global: switch to x/sys/unix

suihkulokki:
The syscall package is locked since 1.4[1], all new changes are in x/sys[2]
The fixes for epoll/arm64 are going to be loaded to x/sys/unix. This
commit is straightforward search/replace with a couple of hand edits.

This is needed for fixing #130

[1] https://golang.org/s/go1.4-syscall
[2] https://godoc.org/golang.org/x/sys/unix

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -13,7 +13,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
-	"syscall"
+	"golang.org/x/sys/unix"
 	"time"
 )
 
@@ -113,12 +113,12 @@ func (w *Watcher) Remove(name string) error {
 		return fmt.Errorf("can't remove non-existent kevent watch for: %s", name)
 	}
 
-	const registerRemove = syscall.EV_DELETE
+	const registerRemove = unix.EV_DELETE
 	if err := register(w.kq, []int{watchfd}, registerRemove, 0); err != nil {
 		return err
 	}
 
-	syscall.Close(watchfd)
+	unix.Close(watchfd)
 
 	w.mu.Lock()
 	isDir := w.paths[watchfd].isDir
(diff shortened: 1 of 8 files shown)


suihkulokki:
Hi, I don't know why the darwin case failed on travis. It appears syscall and x/sys are quite identical in this front.

nathany:
Thanks. It would've been a bit easier to use an alias for the import `import syscall "golang.org/x/sys/unix"` but this is fine.

Adding a dependency could break people's builds, so we'll need to release this carefully.

nathany:
@anthonyfok offered to do some additional testing in #112.

I think we should release this as v1.3.0 by the end of April. I'm thinking of pre-announcing the release to golang-nuts, since it does require a new dependency.

nathany:
Announced: https://groups.google.com/d/msg/golang-nuts/3JLKL785bTY/EacbfZVeAAAJ

kqueue.go
```diff
@@ -23,7 +23,7 @@ type Watcher struct {
 	Errors chan error
 	done   chan bool // Channel for sending a "quit message" to the reader goroutine
 
-	kq int // File descriptor (as returned by the kqueue() syscall).
+	kq int // File descriptor (as returned by the kqueue() unix..
```
anthonyfok: One too many "syscall" string is changed to "unix" here in this comment?  :grin:

suihkulokki:
Updated patch with the comment fixes, one noticed by @anthonyfok thanks. There was another spurious syscall -> unix change + one comment line had become unaligned afterwards.

anthonyfok:
Thank you for updating the patch so quickly, @suihkulokki!

I am happy to report that your patch works wonder!  `hugo server --watch` quickly reports any changed files and re-render the website promptly, and no longer give "function not implemented" errors (spf13/hugo#1772).  Thank you for your great work, @suihkulokki!

Cc: @nathany

nathany:
merged. thanks

PR #145, opened 2016-05-20
Adding test for failure in kqueue close

nikunjy:
Run the test on a mac 
`go test -run TestWatcherClose -count 5`

### DIFF
--- a/kqueue_test.go
+++ b/kqueue_test.go
@@ -1,0 +1,66 @@
+package fsnotify
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"sync"
+	"testing"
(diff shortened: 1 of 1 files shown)


nathany:
Sorry, we need to figure out why these tests are failing.

nathany:
Thanks for providing a failing test. I won't be merging it until we have a fix, nor will I be merging in with an additional testing dependency (as good as testify may be).

nikunjy:
@nathany fair enough. I slacked off on making the fix. Will bounce back with the fix and do the test without testify.

nathany:
Thanks. Would you mind also taking a peek at #124?

PR #147, opened 2016-05-20
Use older syscalls to support older systems (i.e. Redhat 5.6)

jeffwilliams:
I ran into problems compiling the code as it used EpollCreate1 and Pipe2.

### DIFF
--- a/inotify_poller.go
+++ b/inotify_poller.go
@@ -40,12 +40,21 @@ func newFdPoller(fd int) (*fdPoller, error) {
 	poller.fd = fd
 
 	// Create epoll fd
-	poller.epfd, errno = unix.EpollCreate1(0)
+	poller.epfd, errno = unix.EpollCreate(100)
 	if poller.epfd == -1 {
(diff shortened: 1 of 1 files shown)


suihkulokki:
At the moment fsnotify requirement is 2.6.27+ so Redhat 5.6 has probably something older? This PR will break arm64, unless an EpollCreate -> EpollCreate1 wrapper to x/sys is added, since arm64 doesn't have the legacy syscall.

jeffwilliams:
Ah! I didn't realize this would cause an incompatibilty with arm64. It's probably more important to support that than an old kernel. Should I close the request?

nathany:
@jeffwilliams You could checkout an older version of fsnotify and vendor that code. The change to EpollCreate1 was made in v1.2.5. See [releases](https://github.com/fsnotify/fsnotify/releases).

I'm not sure if there is a good way to fallback to the old EpollCreate, so I'm going to go ahead and close this.

PR #149, opened 2016-06-05
Properly handle inotify's IN_Q_OVERFLOW event

zeldovich:
Upon receiving an event with IN_Q_OVERFLOW set in the mask, generate an
error on the Errors chan, so that the application can take appropriate
action.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -245,6 +245,15 @@ func (w *Watcher) readEvents() {
 
 			mask := uint32(raw.Mask)
 			nameLen := uint32(raw.Len)
+
+			if mask&unix.IN_Q_OVERFLOW != 0 {
+				select {
+				case w.Errors <- errors.New("inotify queue overflow"):
+				case <-w.done:
+					return
+				}
+			}
+
 			// If the event happened to the watched directory or the watched file, the kernel
 			// doesn't append the filename to the event, but we would like to always fill the
 			// the "Name" field with a valid filename. We retrieve the path of the watch from
--- a/inotify.go
+++ b/inotify.go
@@ -248,7 +248,7 @@ func (w *Watcher) readEvents() {
 
 			if mask&unix.IN_Q_OVERFLOW != 0 {
 				select {
-				case w.Errors <- errors.New("inotify queue overflow"):
+				case w.Errors <- ErrEventOverflow:
 				case <-w.done:
 					return
 				}
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -9,6 +9,7 @@ package fsnotify
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 )
 
@@ -60,3 +61,6 @@ func (op Op) String() string {
 func (e Event) String() string {
 	return fmt.Sprintf("%q: %s", e.Name, e.Op.String())
 }
+
+// Common errors that can be reported by a watcher
+var ErrEventOverflow = errors.New("fsnotify queue overflow")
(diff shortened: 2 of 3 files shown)


inotify.go
```diff
@@ -245,6 +245,15 @@ func (w *Watcher) readEvents() {
 
 			mask := uint32(raw.Mask)
 			nameLen := uint32(raw.Len)
+
+			if mask&unix.IN_Q_OVERFLOW != 0 {
+				select {
+				case w.Errors <- errors.New("inotify queue overflow"):
```
tsuna: This needs to be a well-defined error otherwise it's not easy for client code to determine that this condition has occurred without inspecting the error string (gasp!).

How about something like:

```go
  var ErrEventOverflow = errors.New("fsnotify queue overflow")
```

(this needs to be placed in a platform-independent file)

so that the client code than check `if err == fsnotify.ErrEventOverflow {`

nathany:
> "IN_Q_OVERFLOW is emitted when the appropriate queue is full and can't take more events. If this event occurs frequently you should read events faster and eventually increase the limit for queues. This limit can be read and changed by reading/changing the appropriate procfs file (/proc/sys/fs/inotify/max_queued_events)." - http://inotify.aiken.cz/?section=inotify&page=faq

Would it be possible to test this by temporarily modifying /proc/sys/fs/inotify/max_queued_events to a small amount? (An inotify-specific tast with the linux build tag)

inotify.go
nathany: Agreed that we should change this. There is no precedent in the code right now, which contains several other errors that are just strings. I would be willing to merge this as is and improve error handling as a followup pull request.

I do think that the error should report that it's the inotify queue that overflowed, as it isn't a queue that's part of fsnotify that had a problem.

zeldovich: Just FYI, the Windows code already has its own special way of handling overflows (see sysFSQOVERFLOW).  It might be worth unifying handling of overflow on Linux and Windows.  That said, I'm a bit swamped in the next week or two, so I'm unlikely to revise this patch until next month at the earliest.  It might make sense to merge this as-is (since it seems no worse than the current behavior), and perhaps unify overflow handling as a separate patch later.

zeldovich:
I suspect that it's possible to test this behavior without resorting to changing `/proc/sys/fs/inotify/max_queued_events`; at least in my experience (which led to discovering and fixing this problem), it's quite easy to generate inotify events faster than fsnotify consumes them.

nathany:
Ok.

We have a new pull request template: https://github.com/fsnotify/fsnotify/blob/master/.github/PULL_REQUEST_TEMPLATE.md

I'd like to ensure all these questions are answered for any pull request.

What does this pull request do?

> Upon receiving an event with IN_Q_OVERFLOW set in the mask, generate an
error on the Errors chan, so that the application can take appropriate
action.

Where should the reviewer start?

How should this be manually tested?

zeldovich:
I've added a well-defined error, `ErrEventOverflow`, which is reported upon encountering inotify queue overflow, and also added a test that checks that `ErrEventOverflow` is properly reported.

As I already mentioned, Windows already has its own special handling for queue overflows, without a well-defined error code.  I did not change how Windows handles overflows as part of this pull request; it would make sense for this to be fixed at some later point (to also return `ErrEventOverflow`).

nathany:
@zeldovich Have you signed the Google CLA? We need contributors to sign the CLA so that fsnotify can be used in Go tools in the future.

https://cla.developers.google.com/about

zeldovich:
Yes, that page says I have a signed individual CLA as of Feb 05, 2013 09:47 PST.

PR #153, opened 2016-06-28
Use filepath.Join()

### LINKED ISSUE #151: One backslash too much when watching files from a root directory
On Windows.

If I watch "c:\", the event name will be "c:\\\\something" instead of "c:\something".

If I watch a directory like "c:\test", the event name is correctly "c:\test\something".

brunoqc:
fix #151

I only tested on Windows. 3 of the 4 changes are covered by tests. Not sure if I should had a test for the fourth one or not.

I think I signed the CLA a while ago.

### DIFF
--- a/windows.go
+++ b/windows.go
@@ -306,7 +306,7 @@ func (w *Watcher) remWatch(pathname string) error {
 		watch.mask = 0
 	} else {
 		name := filepath.Base(pathname)
-		w.sendEvent(watch.path+"\\"+name, watch.names[name]&sysFSIGNORED)
+		w.sendEvent(filepath.Join(watch.path, name), watch.names[name]&sysFSIGNORED)
 		delete(watch.names, name)
 	}
 	return w.startRead(watch)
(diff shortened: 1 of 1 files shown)


nathany:
looks good. thanks

nathany:
Something is flaky with Mac tests, but the Windows tests on Appveyor passed.

Notice that it's logging events with `\\` double slashes everywhere though. Hm.

brunoqc:
> Something is flaky with Mac tests

Isn't `windows.go` only used on Windows?

> Notice that it's logging events with \\\\ double slashes everywhere though. Hm.

Do you see \\\\ double slashes with %s or %q? It should be double with %q I think but not with %s.

nathany:
Yah, don't worry about the the Mac tests. That's a separate issue.

You can see the output in Appveyor (Windows CI) by clicking the link: https://ci.appveyor.com/project/NathanYoungman/fsnotify/build/1.0.209

It may just be a %q vs. %s thing. I'll double check before merging/releasing this.

brunoqc:
I just tested and it seems fine.

```go
package main

import (
	"log"

)

func main() {
	if err != nil {
		panic(err)
	}
	defer watcher.Close()

	go func() {
		for {
			select {
			case event := <-watcher.Events:
				log.Printf("event: %q %s", event.Name, event.Name)
			case err := <-watcher.Errors:
				log.Println("error:", err)
			}
		}
	}()

	err = watcher.Add("D:\\")
	if err != nil {
		panic(err)
	}
	select {}
}
```

output
```
2016/06/28 12:38:10 event: "D:\\Nouvelle image bitmap2.bmp" D:\Nouvelle image bitmap2.bmp
```

nathany:
Merged and tagged. I still have to tag and do a release.

Thanks Bruno

PR #158, opened 2016-08-02
Update docs: relative -> absolute path

jonasfj:
Tested on Unbutu 16.04, maybe it's a relative path on other platforms...

I was monitoring: `/tmp/EFlmKJ8XTYmnxQIucRkY2A/"` and got the event `"/tmp/EFlmKJ8XTYmnxQIucRkY2A/qmp.sock": CREATE`.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -14,7 +14,7 @@ import (
 
 // Event represents a single file system notification.
 type Event struct {
-	Name string // Relative path to the file or directory.
+	Name string // Absolute path to the file or directory.
 	Op   Op     // File operation that triggered the event.
 }
 


nathany:
Maybe the terminology is confusing? The path on Linux should be relative to whatever you are monitoring. If the current working directory was `tmp` and you monitored `EFlmKJ8XTYmnxQIucRkY2A/` then you should see `EFlmKJ8XTYmnxQIucRkY2A/qmp.sock` (relative to tmp).

Whether the filename is included or not doesn't make the path absolute rather than relative.

If you want to strip off qmp.sock for some reason, you can always use `filepath.Dir()` from the standard library.

jonasfj:
>  The path on Linux should be relative to whatever you are monitoring. 

Okay, we have "current working directory" and "the folder being monitored", so relative to what?
Why would I ever want it relative to the current working folder... that's a surprise indeed...
I guess I could make it absolute with `filepath.Abs()`, but being relative to current working folder is not what people would expect, that should be made clear in the docs...

I would expect the `name` to be relative to the folder currently being monitored. I realize in hindsight that this would be unfortunate if I were monitoring multiple folders :)
So it makes a bit sense that it's relative to current working folder, but this seems like a footgun...

Note: I'm monitoring the folder `EFlmKJ8XTYmnxQIucRkY2A` waiting for `qmp.sock` to appear...
So I want to compare when events happen to see if the event was the `qmp.sock` file being created.
As for current working directory, I don't care what that is when my program is running...

I guess the trick is to use `filepath.Abs` to make sure it's absolute...

nathany:
It is relative to the folder being monitored.

If you monitor `/home/go/src` then the path will include `/home/go/src/...`. If you monitor `src` then the path will only include `src/...`. In order to do the later, your current working directory matters.

Not saying it's the best thing. Maybe the paths returned should always be absolute. It would probably work to just call `filepath.Abs` on the path you want to monitor rather than every path returned for a notification -- though we'd need to verify it across each platform.

jonasfj:
> It is relative to the folder being monitored.

What not what you said:
> If the current working directory was tmp and you monitored EFlmKJ8XTYmnxQIucRkY2A/ then you should see EFlmKJ8XTYmnxQIucRkY2A/qmp.sock (relative to tmp).

> If you monitor /home/go/src then the notification paths will include /home/go/src/.... If you monitor src then the path will only include src/.... In order to do the later, your current working directory matters.

Oh, I see... So it's **not a relative path**, it is the path given as input + the path relative to that path...
That's not easy to express... :)

Note: If it was relative to the path being monitored, then in your example above it would be: `test-file` when `test-file` is created in a folder `/home/go/src` regardless of whether or not the input is `src/...` or `/home/go/src`... So it's not relative to the folder being monitored.

This is a hard one... to fix with documentation...
Either we make it absolute, or we have to write a section of docs on the property explaining this, possibly using examples...

nathany:
I opened an issue for this (#160). Thanks for mentioning it.

PR #165, opened 2016-09-14
Add a String() func to Event.Op

oozie:
As a user of the library I want to use `Op.String()` to classify events by their op string without the need to parse `event.String()` or creating my own string builder.

The change delegates string representation of the event to the `Op` receiver, as suggested by a few people before.

Tested on Linux and Darwin.
`
$ go test
PASS
ok  	github.com/fsnotify/fsnotify	13.581s
`
Signed the CLA.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -30,33 +30,33 @@ const (
 	Chmod
 )
 
-// String returns a string representation of the event in the form
-// "file: REMOVE|WRITE|..."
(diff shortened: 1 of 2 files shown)


nathany:
@ishustava regarding ginkgo, see https://github.com/fsnotify/fsnotify/issues/167 for the discussion. thanks.

Issue #164, opened 2016-09-04
fatal error: concurrent map read and map write


jiangytcn:
Is it a bug or something?
After one week or so , my program stop running, only left the exceptions below.
What i want to do is wathcing the changes of the base directory, and if theres sub directory creation event occuied, then add it to the watcher. 

```
fatal error: concurrent map read and map write
goroutine 6 [running]:
runtime.throw(0xab67e0, 0x21)
        /usr/local/go/src/runtime/panic.go:547 +0x90 fp=0xc820195d18 sp=0xc820195d00
runtime.mapaccess1_fast64(0x8b2960, 0xc82000fad0, 0x1, 0xc820195d98)
        /usr/local/go/src/runtime/hashmap_fast.go:112 +0x5a fp=0xc820195d38 sp=0xc820195d18
github.com/XXXXXXXX/gosync/vendor/github.com/fsnotify/fsnotify.(*Watcher).readEvents(0xc820018aa0)
        /home/ubuntu/go/src/github.com/XXXXXXXX/gosync/vendor/github.com/fsnotify/fsnotify/inotify.go:253 +0x6bf fp=0xc8201a5f58 sp=0xc820195d38
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1998 +0x1 fp=0xc8201a5f60 sp=0xc8201a5f58
created by github.com/XXXXXXXX/gosync/vendor/github.com/fsnotify/fsnotify.NewWatcher
        /home/ubuntu/go/src/github.com/XXXXXXXX/gosync/vendor/github.com/fsnotify/fsnotify/inotify.go:61 +0x3e2
```

nathany:
Thanks for the report @jiangytcn.

We are running the tests with `-race` yet it hasn't caught this issue. From your description, it sounds like it rarely occurs, which is going to make it hard to track down.

Please let us known what `go version` you are using and if possible, include some sample code for how you are using fsnotify. Thanks.

oozie:
@jiangytcn 

Is it possible that somewhere in your consuming code you instantiate a watcher, and assign it to a field on a struct by value, as opposed to a reference? In such case more than one mutexes are created to guard the same set of maps.

oozie:
Let me illustrate what I mean with code. The following snippet will fail pretty reliably:
```
package main

import (
	"fmt"
	"io/ioutil"
	"os"
	"path"
	"strconv"
	"time"

	"github.com/fsnotify/fsnotify"
)

[...]

PR #162, opened 2016-08-16
ci: update for Go 1.8

nathany:
two most recent stable versions + tip

PR #169, opened 2016-09-23
Fix creation of multiple mutexes guarding the same set of maps

oozie:
This PR addresses a problem described in https://github.com/fsnotify/fsnotify/issues/164 whereby multiple mutexes are created to guard the same set of maps, which results in concurrent access.

To see multiple mutexes with your own eyes, 
- add print statements to `inotify.go` [like this](https://gist.github.com/oozie/ec901e71053c07f86cd7c6939d1c4485).
- Run the attached integration test on the buggy code. You should see something like this:

```
Locking 0xc4200a00b0   <--- one mutex
Unlocking 0xc4200a00b0
Locking 0xc4200a00b0
Unlocking 0xc4200a00b0
Locking 0xc4200a00b0
Unlocking 0xc4200a00b0
Locking 0xc4200a0060   <---- another mutex
Unlocking 0xc4200a0060
Locking 0xc4200a0060
Unlocking 0xc4200a0060
Locking 0xc4200a00b0
Locking 0xc4200a0060
fatal error: concurrent map read and map write

```
- Healthy code should produce only one address:

```
$ go test -run Mutex
Start!
...
Unlocking 0xc4200921a0
Locking 0xc4200921a0
Unlocking 0xc4200921a0 later
Locking 0xc4200921a0
Unlocking 0xc4200921a0
Locking 0xc4200921a0
Unlocking 0xc4200921a0 later
Locking 0xc4200921a0
Unlocking 0xc4200921a0
Locking 0xc4200921a0
Unlocking 0xc4200921a0 later
Locking 0xc4200921a0
PASS
ok      github.com/fsnotify/fsnotify    0.104s
$
```

Tested on Darwin and Linux. 
CLA signed.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -23,8 +23,8 @@ import (
 type Watcher struct {
 	Events   chan Event
 	Errors   chan error
-	mu       sync.Mutex // Map access
-	cv       *sync.Cond // sync removing on rm_watch with IN_IGNORE
+	mu       *sync.Mutex // Map access
+	cv       *sync.Cond  // sync removing on rm_watch with IN_IGNORE
 	fd       int
 	poller   *fdPoller
 	watches  map[string]*watch // Map of inotify watches (key: path)
@@ -55,8 +55,9 @@ func NewWatcher() (*Watcher, error) {
 		Errors:   make(chan error),
 		done:     make(chan struct{}),
 		doneResp: make(chan struct{}),
+		mu:       &sync.Mutex{},
 	}
-	w.cv = sync.NewCond(&w.mu)
+	w.cv = sync.NewCond(w.mu)
 
 	go w.readEvents()
 	return w, nil
--- a/integration_test.go
+++ b/integration_test.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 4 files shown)


oozie:
@nathany Just a side note: this is not `inotify`-specific. The issue is reproducible on Darwin with `kqueue`, I just instructed the curious tester fit print statements in `inotify`, but the same can be done in `kqueue.go`.

nathany:
Thanks @oozie. Still need to deal with flaky kqueue tests in #162 also -- not sure if that's related.

I've not seen a pointer to a Mutex in common practice, so this is interesting. 

Do you think it would be worth restructuring the code further to make it clear what the mutex guards? I really like this idea: https://talks.golang.org/2015/tricks.slide#15

twpayne:
If you embed a `sync.Mutex` value in a struct then that `sync.Mutex`'s value will be copied when the struct is copied. This is almost certainly not what you want. The compiler doesn't catch this, but `go vet` does.

Since `Watcher` values should *not* be copied, it should be OK to be embed a `sync.Mutex` value (not pointer) in a `Watcher`.

In this PR, the test explicitly copies `Watcher` objects which is why @oozie  sees multiple `sync.Mutex`s in the output.

#189 likely fixes the underlying problems in the library. @oozie, you should run `go vet` on your code to make sure that you're not copying `sync.Mutex` values.

matthias-stone:
@oozie, Watchers shouldn't be copied by value (as they map to a single set of kernel resources), and go vet will catch this.

Closing in agreement with @twpayne's analysis.

PR #174, opened 2016-10-05
create inotify fd with close-on-exec + deflake inotify stress test

pattyshack:
(on behalf of @msolo) inotify fd should not leak to children processes (this matches the kqueue implementation behavior).

Issue #150, opened 2016-06-05
Data race in test on tip/Linux


bep:
Since tip is now 1.7beta we should look into this and report any real issue to the Go people:

https://travis-ci.org/fsnotify/fsnotify/jobs/135335818
https://travis-ci.org/fsnotify/fsnotify/jobs/124355194

I have not looked closely into this, but I tried to rebuild it once with the same error.

bep:
OK, it is established that this is a faulty test.

nathany:
Thanks @bep. It seems all the current pull requests are failing in odd ways. We should really look into fixing or removing faulty tests.

nathany:
This may be related to #142.

> panic: Fail in goroutine after TestInotifyStress has completed

Issue #117, opened 2016-02-14
Events are not triggered when running in background


thinkclay:
When using nohup or & to suppress a process that leverages fsnotify, the events are no longer received in the background process. Implementation here: https://github.com/wellington/wellington/blob/master/filewatcher.go#L149

nathany:
Which platform are you running into this issue on?

thinkclay:
Mac: 10.11.4 Beta (15E27e)

nathany:
Thanks. 

Were the events happening previously? Do you have any idea of what may have changed in wellington, or if it was an update to fsnotify?

drewwells:
I'd have to build this again on older branches, if you have a guess when this could be a regression that would help track it down. I probably never have tested this as a background process, so not sure if it's a regression or always been a problem.

arp242:
I can't seem to reproduce this; on macOS:

	[~/fsnotify]% go run ./cmd/fsnotify /tmp/watch/ &
	[1] 1130
	[~/fsnotify]% 00:86:46.9177 watching; press ^C to exit
	00:86:49.9164   1 CREATE                "/tmp/watch/asd"
	00:86:52.1444   2 CREATE                "/tmp/watch/asdzxczxc"
	00:86:57.6487   3 CREATE                "/tmp/watch/qwewqezxc"

	[~/fsnotify]% fg
	[1]  + running    go run ./cmd/fsnotify /tmp/watch/
	^Csignal: interrupt

And with nohup:

	1[~/fsnotify]% nohup go run ./cmd/fsnotify /tmp/watch/ &
	[1] 1151
	[~/fsnotify]% appending output to nohup.out

	[~/fsnotify]% cat nohup.out
	00:86:18.8525 watching; press ^C to exit
	00:86:26.0873   1 CREATE                "/tmp/watch/NOHUP"
	00:86:26.8113   2 CREATE                "/tmp/watch/NOHUP2"
	[~/fsnotify]% fg
	[1]  + running    nohup go run ./cmd/fsnotify /tmp/watch/
	^C

Since this is an old issue that I can't reproduce, I'll close this for now; please get back to me if you're still experiencing it and we can reopen and investigate further.

PR #177, opened 2016-10-05
deflake inotify stress test

### LINKED ISSUE #142: Flaky test TestInotifyStress
Hi again,

While testing the changes to fix #130 in arm64,  I noticed that sometimes (specially when the host is very loaded) the TestInotifyStress test would fail.

I can reproduce this consistently if I raise CPU usage artificially and run the tests with nice.

```
=== RUN   TestInotifyStress
--- FAIL: TestInotifyStress (5.02s)
	inotify_test.go:236: Creates and removes should not be off by more than one: 287 creates, 291 removes
```

pattyshack:
this addresses the comments in ~~#117~~ #174:
  - the errChan buffer is necessary to ensure the file generation goroutine does not get block by the main test goroutine
  - got rid of the sleeps in the file generation goroutine (reduce test time by a few seconds on my laptop)
  - create files based on a fixed number instead of a fixed amount of time
  
edit: fixes #142 #150

### DIFF
--- a/inotify_test.go
+++ b/inotify_test.go
@@ -174,7 +174,7 @@ func TestInotifyStress(t *testing.T) {
 	}
 
 	doneChan := make(chan struct{})
-	// The buffer ensures that the file generation goroutine is never	blocked.
+	// The buffer ensures that the file generation goroutine is never blocked.
 	errChan := make(chan error, 2*maxNumToCreate)
 
 	go func() {
--- a/inotify_test.go
+++ b/inotify_test.go
@@ -194,6 +194,10 @@ func TestInotifyStress(t *testing.T) {
 			}
 		}
 
+		// If we delete a newly created file too quickly, inotify will skip the
+		// create event and only send the delete event.
+		time.Sleep(100 * time.Millisecond)
+
 		for i := 0; i < maxNumToCreate; i++ {
 			testFile := fmt.Sprintf("%s%d", testFilePrefix, i)
 			err = os.Remove(testFile)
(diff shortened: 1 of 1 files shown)


nathany:
It still seems to be a bit flaky (#142) for me, though the data race (#150) is resolved.

```
vagrant@vagrant-ubuntu-trusty-64:~/src/github.com/fsnotify/fsnotify$ go test ./... -race
--- FAIL: TestInotifyStress (1.43s)
	inotify_test.go:263: Creates and removes should not be off by more than one: 2 creates, 1000 removes
FAIL
FAIL	github.com/fsnotify/fsnotify	9.945s
vagrant@vagrant-ubuntu-trusty-64:~/src/github.com/fsnotify/fsnotify$ go test ./... -race
ok  	github.com/fsnotify/fsnotify	10.738s
vagrant@vagrant-ubuntu-trusty-64:~/src/github.com/fsnotify/fsnotify$ 
```

pattyshack:
i guess the sleep is necessary after all.  basically the problem is if we delete a newly created file too quickly, the OS will skip the create events and only send the delete event.  (looks like integration test is flaky as well =( )

nathany:
Yah, looks that way. https://github.com/fsnotify/fsnotify/issues/142#issuecomment-242377927

Thanks for your work on this.

pattyshack:
$ go test -test.count 10 .
ok  	github.com/fsnotify/fsnotify	88.140s

nathany:
Thanks, it's working well for me also.

FYI, this stress test was added in #66.

nathany:
Thanks. P.S. I invited you to the repo, but please continue to use branches and pull requests for any code changes.

pattyshack:
Cool.  Will do

Issue #166, opened 2016-09-14
concurrent map read and map write on printing Watcher struct


oozie:
This is a relatively minor problem. I found it when playing with https://github.com/fsnotify/fsnotify/issues/164

The `%v` serializer accesses watcher's maps without a lock on `fmt.Println`. Run the following

``` go
package main

import (
    "fmt"
    "os"
    "path"
    "strconv"
    "time"

    "github.com/fsnotify/fsnotify"
)

func main() {

    testDirPrefix := "/tmp/test"

    w, _ := fsnotify.NewWatcher()

    os.MkdirAll(testDirPrefix, 0700)

    for i := 0; i < 2; i++ {
        dirName := path.Join(testDirPrefix, strconv.Itoa(i))
        os.MkdirAll(dirName+"/subdir", 700)
        go func() {
            for {
                w.Add(dirName + "/subdir")
                fmt.Println(w)
                w.Remove(dirName + "/subdir")
            }
        }()
    }

    time.Sleep(time.Second)

    os.RemoveAll(testDirPrefix)
}
```

and you shall see 

``` go
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/0/subdir:true /tmp/test/1/subdir:true] map[] map[] map[] false}
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/1/subdir:true /tmp/test/0/subdir:true] map[] map[] map[] false}
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/1/subdir:true /tmp/test/0/subdir:true] map[] map[] map[] false}
fatal error: concurrent map read and map write
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/0/subdir:true /tmp/test/1/subdir:true] map[] map[] map[] false}
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/0/subdir:true /tmp/test/1/subdir:true] map[] map[] map[] false}
&{0xc42006e060 0xc42006e0c0 0xc42006e120 3 {0 0} map[] map[/tmp/test/0/subdir:true /tmp/test/1/subdir:true] map[] map[] map[] false}

goroutine 19 [running]:
runtime.throw(0xccff1, 0x21)
[...]

PR #180, opened 2016-10-13
Prevent concurrent map access when inspecting `Watcher`

oozie:

This PR addresses https://github.com/fsnotify/fsnotify/issues/166 where inspecting of `Watcher` may lead to concurrent map access.

#166 is a good place to start. First commit is a breaking test that illustrates the problem through an example. The following commit is the fix.

use `fmt.Println` to print a watcher instance while `Add` and `Remove` calls are issued to the watcher.

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -7,6 +7,7 @@
 package fsnotify
 
 import (
+	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


oozie:
Backing this PR out for now.

PR #196, opened 2017-02-05
Add support for solaris FEN

### LINKED ISSUE #12: File Events Notifications on Solaris
@4ad would like fsnotify to support FEN on Solaris.

While each file & directory requires an event source, the [limits are much higher than kqueue](https://groups.google.com/d/msg/golang-dev/bShm2sqbrTY/HMm7Y36uuxIJ):

> The default limit is 64k event sources per event port per process.
> The default limit for the number of event sources per system is
> 2^31.
> 
> The default limit for the number of event ports per process is 8k.
> The default limit for the number of event ports per system is 64k.
### Reference
- [File Events Notification](https://blogs.oracle.com/praks/entry/file_events_notification)
- DTrace on Solaris can watch recursively.

gfrey:
Fixes #12
I've signed the CLA.

This is a basic implementation of fsnotify for Solaris FEN. It at least passes all integration tests.

I used the following two resource to get a basic understanding how this should work:
* https://solarisrants.wordpress.com/2013/07/24/solaris-file-event-notification/
* https://blogs.oracle.com/praks/entry/file_events_notification

The available integration tests are enabled for solaris now and succeed.

### DIFF
--- a/fen.go
+++ b/fen.go
@@ -6,32 +6,274 @@
 
 package fsnotify
 
+// #include <errno.h>
+// #include <strings.h>
+// #include <unistd.h>
+// #include <stdio.h>
+// #include <port.h>
+// #include <sys/stat.h>
+//
+// uintptr_t from_file_obj(struct file_obj *obj) {
+//   return (uintptr_t)obj;
+// }
+//
+// struct file_obj*to_file_obj(uintptr_t ptr) {
+//   return (struct file_obj*)ptr;
+// }
+//
+// struct file_info {
+//   uint mode;
+// };
+import "C"
 import (
 	"errors"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"sync"
+	"syscall"
+	"unsafe"
 )
 
 // Watcher watches a set of files, delivering events to a channel.
 type Watcher struct {
 	Events chan Event
 	Errors chan error
+
+	port C.int
+	wg   *sync.WaitGroup
+
+	mu      sync.Mutex
+	watches map[string]*C.struct_file_obj
+
+	closed bool
 }
 
-// NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
 func NewWatcher() (*Watcher, error) {
-	return nil, errors.New("FEN based watcher not yet supported for fsnotify\n")
+	var err error
+
+	w := new(Watcher)
+	w.Events = make(chan Event)
+	w.Errors = make(chan error)
+	w.watches = make(map[string]*C.struct_file_obj)
+
+	w.port, err = C.port_create()
+	if err != nil {
+		return nil, err
+	}
+
+	w.wg = new(sync.WaitGroup)
+	w.wg.Add(1)
+	go w.run()
+
+	return w, nil
 }
 
-// Close removes all watches and closes the events channel.
 func (w *Watcher) Close() error {
+	if w.closed {
+		return nil
+	}
+	w.closed = true
+
+	C.close(w.port)
+	w.wg.Wait()
+
+	close(w.Events)
+	close(w.Errors)
 	return nil
 }
 
-// Add starts watching the named file or directory (non-recursively).
-func (w *Watcher) Add(name string) error {
+func (w *Watcher) Add(path string) error {
+	stat, err := os.Stat(path)
+	switch {
+	case err != nil:
+		return err
+	case stat.IsDir():
+		return w.handleDirectory(path, stat, w.associateFile)
+	default:
+		return w.associateFile(path, stat)
+	}
+}
+
+func (w *Watcher) Remove(path string) error {
+	if !w.watched(path) {
+		return nil
+	}
+
+	stat, err := os.Stat(path)
+	switch {
+	case err != nil:
+		return err
+	case stat.IsDir():
+		return w.handleDirectory(path, stat, w.dissociateFile)
+	default:
+		return w.dissociateFile(path, stat)
+	}
+}
+
+func (w *Watcher) run() {
+	defer w.wg.Done()
+
+	for {
+		var pevent C.port_event_t
+		_, err := C.port_get(w.port, &pevent, nil)
+		if err != nil {
+			if err.(syscall.Errno) == C.EBADF {
+				return
+			}
+			w.Errors <- err
+		}
+
+		if pevent.portev_source != C.PORT_SOURCE_FILE {
+			// Event from unexpected source received; should never happen.
+			continue
+		}
+
+		finfo := (*C.struct_file_info)(pevent.portev_user)
+		err = w.handleEvent(pevent.portev_object, pevent.portev_events, finfo)
+		if err != nil {
+			w.Errors <- err
+		}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


fen.go
```diff
@@ -231,7 +231,7 @@ func (w *Watcher) associateFile(path string, stat os.FileInfo) error {
 	var finfo C.struct_file_info
 	finfo.mode = C.uint(stat.Mode())
 
-	var mode C.int = C.FILE_MODIFIED | C.FILE_ATTRIB | C.FILE_NOFOLLOW
+	mode := C.FILE_MODIFIED | C.FILE_ATTRIB | C.FILE_NOFOLLOW
```
gfrey: Well type inference doesn't seem to work as houndci-bot expected. Now I need to cast the usage of mode. Not sure I like that better. Will have a closer look later.

nathany:
@4ad @amitkris Please review.

nathany:
@gfrey We don't currently have a good way to test Solaris, either with a CI service or locally from a different OS (I use [Vagrant Gopher](https://github.com/nathany/vagrant-gopher) for Linux/BSD). Any suggestions?

gfrey:
@nathany I used VirtualBox with an OpenIndiana hipster install. I'll try and add support for solaris to your Vagrant Gopher, just give me a few days.

nathany:
@gfrey Thanks both for this pull request and your work on Vagrant Gopher to test it.

```
❯ vagrant ssh solaris -c 'go version; cd fsnotify/fsnotify; go test ./...'
go version go1.8 solaris/amd64
/export/home/vagrant/src/github.com/fsnotify/fsnotify
ok      github.com/fsnotify/fsnotify    8.332s
Connection to 127.0.0.1 closed.
```

(unfortunately no race detector on Solaris)

I'm still hoping @4ad and/or @amitkris provide a review, but I will also take a closer look in the coming days.

fazalmajid:
What can we do to get this PR accepted and merged into the master branch?

This is blocking getting https://github.com/spf13/hugo and https://github.com/prometheus/alertmanager (via https://github.com/stuartnelson3/guac) running on Solaris.

I confirmed that this branch merged into master works on SmartOS inside Hugo (running `hugo --watch`) whereas without it it just exits with the message `FEN based watcher not yet supported for fsnotify`. Even if gfrey's code has bugs, it can't be any more broken on Solaris than the stub in the master branch. Furthermore, `fen.go` is marked as `+build solaris` so there is zero risk of regression on other platforms.

bep:
@4ad @amitkris @nathany This would be really good for Hugo to get in. @fazalmajid has a valid point about the _this cannot break it worse than it already is_ and the breakage is limited to Solaris. I have quickly browsed through the code and cannot find any obvious mistakes.

tiwaana:
We in DTC golang meetup reviewed the code, not all it. But we have few questions
a) is using C required
b) In close() things are not mutex protected. 
its only limited to Solaris, so should not break other things.
Anand

binarycrusader:
I'm not the author, but Solaris requires use of cgo, in general, since libc is the provider of system interfaces.  The only other comment I would add is that this should be using golang.org/x/sys/unix instead of syscall.  Per the official Go documentation, use of syscall outside of the Go standard runtime has been deprecated since 1.4:

> NOTE: This package is locked down. Code outside the standard Go repository should be migrated to use the corresponding package in the golang.org/x/sys repository. That is also where updates required by new systems or versions should be applied. See https://golang.org/s/go1.4-syscall for more information. 
https://golang.org/pkg/syscall/

As far as I can tell, this should work on commercial releases of Solaris, although I haven't tried it myself.  It appears to be using functionality shared between Solaris and OpenSolaris-based derivatives such as Illumos/SmartOS.

4ad:
I don't care about fsnotify, so you'll have to find someone else to review this, but I will point out that using cgo here is an unnecessary mistake. This should all be done using golang.org/x/sys/unix (probably necessary to update it).

quenelle:
Forgive any mistakes here, I'm very new to Go.  For this kind of code there seems to be a good justification for using the C pass through interfaces.  Probably lots of things don't have native Go interfaces yet.  Using "syscall" is another matter.  It's explicitly deprecated.  I see two places where syscall is used explicitly in the code.  1) It's used to access the enumeration of errno values.  When I visit x/sys/unix, it says that the official place to get errno values from is syscall.Errno.  (Check the top of https://godoc.org/golang.org/x/sys/unix) So that seems legitimate.  2) The other place it's used is to access the C structure for stat_t.  The code starts by using os.FileInfo, then goes through Sys() to get the stat_t.  It does this because the FileInfo doesn't include atime or ctime members. (access time / create time).  Only mtime (modification time) is available through FileInfo according to the docs.  So I don't see any way the code could be changed to avoid the existing uses of syscall.

In theory you could argue that atime and ctime should be added FileInfo, but that seems like a change that should take into account the availability of those interfaces on other platforms.  I'm not sure how widely they're supported on all the Go platforms.  It doesn't seem appropriate to ask the submitter to take that on at this point.

So with that in mind, I don't see any issues with the use of cgo or syscall.

I also ran the fsnotify package test on x86 Oracle Solaris (latest internal build) with the bundled Go 1.5, and it passed.  I also ran it on the same Solaris build on SPARC with the latest sparc compiler from https://github.com/4ad/go/tree/sparc64.go1.8 and it passed.

I'm not that familiar with file events in Solaris, but if additional code review is what's holding this up, I can find an Oracle Solaris engineer who is familiar with file events to check over the logic.  They won't be familiar with Go, but they could vouch for the usage of Solaris interfaces.

maxmeyer:
@dago Are you still into OpenCSW? Can their build farm help to solve the CI "problem" in this issue?

isaacdavis:
Hi @gfrey! I work at Joyent. We maintain SmartOS, and we need fsnotify support.

Thank you for doing this work! I took a look at your pull request and made some changes. I added more robust support for cancellation to fix a race condition exhibited by the TestWatcherClose test, and I altered the interpretation of FEN events to more accurately implement fsnotify's expected behavior, as well as other small improvements.

I intend to submit a new pull request for this work. Given that my request would still substantially contain your code, how would you like to be credited? I could add you to the `AUTHORS` file or anything else you might prefer. Thanks!

To everyone else on this thread - I would be happy to provide additional resources for testing fsnotify on SmartOS. Please let me know!

gfrey:
@isaacdavis I'm fine with being mentioned in the AUTHORS file. Nice seeing this getting some traction again!

maxmeyer:
@dago What about setting about an automatic build on the opencsw build farm for this project?

dago:
@maxmeyer That would be doable, we already have a [number of projects](https://www.opencsw.org/extend-it/signup/to-upstream-maintainers/) in a [buildbot setup](https://buildfarm.opencsw.org/buildbot/waterfall) which is triggered by hooks, most notable curl which has full CI integration in GitHub with different builders. However, I need to update some stuff to recent Go compilers and make it compile manually first. Please PM me for details.

maxmeyer:
@isaacdavis can you help @dago with details to make it build automatically for csw?

nathany:
Closed in favour of #263.

nathany:
See #371

Issue #191, opened 2016-12-07
Need an Operation for Linux - IN_CLOSE_WRITE


radu-munteanu:
The real world scenario I have is about knowing when a file is fully copied to a location via FTP, SMB or other means, on Linux. The existing Create doesn't do fully justice to files, only to directories. The Write (IN_MODIFY) operation is not fully helpful in this case, because I can get a lot of these messages, but not knowing exactly when the transfer has finished.

I don't know if there is something similar to IN_CLOSE_WRITE on other systems, that's why I made this issue, to create the base for a discussion.

One way to solve this problem is to create a new cross-platform operation, which on some systems might have the same meaning with Write.

Another way, that is a bit ugly, but will fix any similar issues in the future, is to add the uint32 Mask to the Event. This way any developer could handle the Mask itself, in case it builds a platform specific software.

radu-munteanu:
I've created a new Op called "Update" on my fork, and used it on inotify to represent  unix.IN_CLOSE_WRITE  and unix.IN_MOVED_TO. I didn't write any tests. 

[https://github.com/radu-munteanu/fsnotify/commit/6638bfdb7939ba8a4687a44cdff21906d9b547c9](https://github.com/radu-munteanu/fsnotify/commit/6638bfdb7939ba8a4687a44cdff21906d9b547c9)

Warning: I didn't write any tests for this Op.

radu-munteanu:
I've created a new Op called "Update" on my fork, and used it on inotify to represent  unix.IN_CLOSE_WRITE  and unix.IN_MOVED_TO. I didn't write any tests. 

[https://github.com/radu-munteanu/fsnotify/commit/a52516bed152ed36854ef441dce07b0c4dfce458](https://github.com/radu-munteanu/fsnotify/commit/a52516bed152ed36854ef441dce07b0c4dfce458)

Warning: I didn't write any tests for this Op.

nathany:
related: #22, #131

hallyn:
Hi, (1) are you planning on doing a pull request from your fork, and (2) would you by any chance consider adding a purer CLOSE (IN_CLOSE_WRITE|IN_CLOSE_NOWRITE) op?

nathany:
I think this needs more discussion before opening any pull requests, as it likely involves an addition or change to the API, and fsnotify is a cross-platform API.

[...]

PR #202, opened 2017-03-28
Added OPEN and CLOSE events

betim:
Adds OPEN and CLOSE events from unix.*

inotify.go and fsnotify.go

Add a folder with watcher.Add(...) then try to open/modify/write file.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -29,6 +29,8 @@ const (
 	Remove
 	Rename
 	Chmod
+	Open
+	Close
 )
 
 func (op Op) String() string {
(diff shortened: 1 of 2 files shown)


radu-munteanu:
You won't get that merge approved, as it only works with one OS. Check this thread: https://github.com/fsnotify/fsnotify/issues/191

betim:
@radu-munteanu, you mean it does not work in one OS?

I totally understand that fsnotify tends to be as cross-platform as possible. But not implementing something rather essential I cannot. Besides windows has file locking so combination of file locking and `Op.Write` event should give windows programmer freedom to implement the same functionality as getting `Op.Close` on linux.

bevand10:
@radu-munteanu - I'd recommend switching to this fork as it'll better suit your needs https://github.com/ericbrisrubio/fsnotify

Dmitry-N-Medvedev:
@bevand10, Dave, if there is no IN_CLOSE_WRITE or similar operation how would one implement processing of newly created or changed files? I personally, being a nub tho, don't know how that could be implemented.

Without CLOSE operation the fsnotify seems to have next to now real-world application at all.

betim:
@bevand10 - I'm using my own fork for this, it works very well.

PR #203, opened 2017-03-29
Fix deadlock in Remove (linux/inotify)

### LINKED ISSUE #115: Not able to remove file watcher
I need to remove file watcher. Calling method Remove(fileName) when event happened to prevent receive other event's to this file.

And actually this call to remove event is not working. It's waiting for ignoreLinux to remove watcher, but this never happening.

### LINKED ISSUE #123: Removing watches from a goroutine consuming fsnotify events can lead to a deadlock
As a result of the fix in fsnotify/fsnotify#73, `watch.Remove()` now synchronously waits for the main fsnotify goroutine (`readEvents()`) to receive and process the `IN_IGNORED` event that the kernel sends to acknowledge the removal of the watch descriptor.  But there could be other inotify events delivered in between, and this can easily cause `readEvents()` to attempt to dispatch an event to someone waiting for `readEvents()` to process the `IN_IGNORED` event, hence the deadlock.

An example of this issue can be found in hpcloud/tail#75 (along with a workaround for that code).

### LINKED ISSUE #195: inotify: Watcher.Remove may deadlock when multiple fs events occur
Before reporting an issue, please ensure you are using the latest release of fsnotify.

Distributor ID: Ubuntu
Description:    Ubuntu 14.04.1 LTS
Release:        14.04
Codename:       trusty
go version:      go1.7.3 linux/amd64

I run into this issue, and from issue list I think people seeming to have similar problem, so I spend some time digging the cause.

The code I am using is latest commit fd9ec7d
Assuming select loop for events is sequential, see code example below 

Inotify may generate multiple events at same time. When this happens, if one happens to call Watcher.Remove on **_any valid watch path_** (e.g. for error handling), then it will block on the Watcher's condition variable (inotify.go:158), waiting for a broadcast on sender side (inotify.go:299). However, the channel size is 1, so it will also block waiting for the select loop to consume events, which is already in waiting state.

I tried to increase the event channel size, which allows sender side to proceed, or alternatively one could use goroutine for Remove in order to avoid the deadlock. But I am not sure if this is ideal.

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.
Code below is to reproduce the issue: a test folder "tests" has a file "file1" inside, and the code tries to rename "tests/file1" to "file2" (which generates two events). If Remove is called inside select loop, it will cause deadlock alarm. Changing ```watcher.Remove``` to ```go watcher.Remove``` or modify watcher's Events channel would fix it.

```go
package main

import (
        "log"
        "os"
        "time"

)

func main() {
        watcher, _ := fsnotify.NewWatcher()
        watcher.Add("tests")
        watcher.Add("tests/file1")

        done := make(chan bool)
        first := true

        go func() {
                for {
                        select {
                        case event := <-watcher.Events:
                                log.Printf("event name: %s", event.String())
                                if first {
                                        // *boom*
                                        watcher.Remove("tests/file1")
                                        first = false
                                }
                                //log.Printf("is this solved?")
                        }
                }
        }()

        time.Sleep(2 * time.Second)
        os.Rename("tests/file1", "file2")
        <-done
}
```

aarondl:
This resolves a fairly common and easy to hit deadlock that occurs in inotify/linux. The issues reported are quoted below.
Commit message follows:

> Several people have reported this issue where if you are using a
> single goroutine to watch for fs events and you call Remove in
> that goroutine it can deadlock. The cause for this is that the Remove
> was made synchronous by PR #73. The reason for this was to try and
> ensure that maps were no longer leaking.
> 
> In this PR: IN_IGNORE was used as the event to ensure map cleanup.
> This worked fine when Remove() was called and the next event was
> IN_IGNORE, but when a different event was received the main goroutine
> that's supposed to be reading from the Events channel would be stuck
> waiting for the sync.Cond, which would never be hit because the select
> would then block waiting for someone to receive the non-IN_IGNORE event
> from the channel so it could proceed to process the IN_IGNORE event that
> was waiting in the queue. Deadlock :)
> 
> Removing the synchronization then created two nasty races where Remove
> followed by Remove would error unnecessarily, and one where Remove
> followed by an Add could result in the maps being cleaned up AFTER the
> Add call which means the inotify watch is active, but our maps don't
> have the values anymore. It then becomes impossible to delete the
> watches via the fsnotify code since it checks it's local data before
> calling InotifyRemove.
> 
> This code attempts to use IN_DELETE_SELF as a means to know when a watch
> was deleted as part of an unlink(). That means that we didn't delete the
> watch via the fsnotify lib and we should clean up our maps since that
> watch no longer exists. This allows us to clean up the maps immediately
> when calling Remove since we no longer try to synchronize cleanup
> using IN_IGNORE as the sync point.
> 
> - Fix #195
> - Fix #123
> - Fix #115

Fixes the deadlocks reported in the quoted issues.

Make a small program that simply watches a directory, and on response to the create event, delete the watch from the directory. Once that's done, run the program, and create a file inside the directory. If you delete and create the file rapidly enough times the deadlock should occur with the old code, but will no longer occur with this code.

I've signed the CLA and REALLY didn't want to. I'm fairly upset that this repo requires this, and then I read this only to find out I signed it without reason: https://github.com/golang/go/issues/4068#issuecomment-250958473
We should remove the requirement from the CONTRIBUTING.md immediately.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -24,7 +24,6 @@ type Watcher struct {
 	Events   chan Event
 	Errors   chan error
 	mu       sync.Mutex // Map access
-	cv       *sync.Cond // sync removing on rm_watch with IN_IGNORE
 	fd       int
 	poller   *fdPoller
 	watches  map[string]*watch // Map of inotify watches (key: path)
@@ -56,7 +55,6 @@ func NewWatcher() (*Watcher, error) {
 		done:     make(chan struct{}),
 		doneResp: make(chan struct{}),
 	}
-	w.cv = sync.NewCond(&w.mu)
 
 	go w.readEvents()
 	return w, nil
@@ -137,6 +135,13 @@ func (w *Watcher) Remove(name string) error {
 	if !ok {
 		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
 	}
+
+	// We successfully removed the watch if InotifyRmWatch doesn't return an
+	// error, we need to clean up our internal state to ensure it matches
+	// inotify's kernel state.
+	delete(w.paths, int(watch.wd))
+	delete(w.watches, name)
+
 	// inotify_rm_watch will return EINVAL if the file has been deleted;
 	// the inotify will already have been removed.
 	// watches and pathes are deleted in ignoreLinux() implicitly and asynchronously
@@ -152,13 +157,6 @@ func (w *Watcher) Remove(name string) error {
 		// Watch descriptors are invalidated when they are removed explicitly or implicitly;
 		// explicitly by inotify_rm_watch, implicitly when the file they are watching is deleted.
 		return errno
-	}
-
-	// wait until ignoreLinux() deleting maps
-	exists := true
-	for exists {
-		w.cv.Wait()
-		_, exists = w.watches[name]
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


nullbio:
An example of a library with problems that this patch fixes: https://github.com/markbates/refresh/issues/6

purpleidea:
@aarondl RE:

> I've signed the CLA and REALLY didn't want to. I'm fairly upset that this repo requires this, and then I read this only to find out I signed it without reason: golang/go#4068 (comment)

I too am strongly against this! +1

Thanks for the patch. Look forward to testing this in https://github.com/purpleidea/mgmt/ shortly.

PR #188, opened 2016-11-12
avoid deadlocks on close. Fixes #187

glycerine:
Raw channel work (not inside a select) is subject to deadlock on shutdown.

Hence Close() can deadlock.

This is already fixed in the linux inotify.go code.

Fixes #187

PR #211, opened 2017-06-27
Small code cleanups for kqueue

tsuna:

Just a couple small code cleanups for the kqueue implementation.

It's a really simple change :)

Tests passed on macOS 10.12.5 including additional tests for my company's internal packages that use fsnotify under the hood.

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -22,7 +22,7 @@ import (
 type Watcher struct {
 	Events chan Event
 	Errors chan error
-	done   chan bool // Channel for sending a "quit message" to the reader goroutine
+	done   chan struct{} // Channel to close to signal to the reader goroutine to exit
 
 	kq int // File descriptor (as returned by the kqueue() syscall).
 
@@ -56,7 +56,7 @@ func NewWatcher() (*Watcher, error) {
 		externalWatches: make(map[string]bool),
 		Events:          make(chan Event),
 		Errors:          make(chan error),
-		done:            make(chan bool),
+		done:            make(chan struct{}),
 	}
 
 	go w.readEvents()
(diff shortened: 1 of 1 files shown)


kqueue.go
```diff
@@ -71,10 +71,8 @@ func (w *Watcher) Close() error {
 		return nil
 	}
 	w.isClosed = true
-	w.mu.Unlock()
 
 	// copy paths to remove while locked
-	w.mu.Lock()
 	var pathsToRemove = make([]string, 0, len(w.watches))
 	for name := range w.watches {
```
omeid: The loop can be simply replaced with `pathsToRemove = append(pathsToRemove, w.watches...)`.

omeid: And I am not sure the make is really needed there, so the whole thing can be `pathsToRemove := append([]string{}, w.watches...)`.

tsuna: I thought the same thing initially but `w.watches` is a map, so we can't `append()` directly, we need to iterate over its keys.

omeid: That makes sense.

tsuna:
The build failed for a seemingly unrelated reason.  Going to close/re-open this PR to make Travis give it another go.

kqueue.go
cpuguy83: Make also ensures you make only one allocation for exactly what you need rather than continuously allocating on append (though append is smart and requests more than is needed to prevent multiple alloc requests)

cpuguy83:
This change actually prevents a race condition on Close. Maybe not super important because I doubt too anyone is calling Close() from multiple goroutines, but still.

cpuguy83:
This looks to be a dup of #188 but w/o fixing some of the other deadlock scenarios.

tsuna:
Oh wow I totally forgot about this change.  Thanks for taking a look at it.  It's not meant to fix anything it was really just a tiny code cleanup.

nathany:
Sorry for leaving this so long. It looks like #230 made these changes and was merged a while ago.

PR #218, opened 2017-08-30
Remove of unexistent is not an error

bo0rsh201:
Technically it's not an error when we are trying to delete unexistent watch. This error message is meaningless.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -8,7 +8,6 @@ package fsnotify
 
 import (
 	"errors"
-	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


cpuguy83:
I think it's a good idea to return an error, but maybe something that the caller can inspect and see that it's specifically a not found kind of error so they can make the determination of how to handle it.

cpuguy83:
For example, see `os.IsNotExist(err)`

bo0rsh201:
Thank you for answer. As for now I see that this remove is 99% useless in my case, because you force remove on DELETE event for directory, so I don't need to call it (maybe in case when we don't watch parent dir, but a couple of separate dirs with different unwatched parents)

Issue #155, opened 2016-07-06
watchers leak inotify file descriptor when process fork/execs a child


msolo-dropbox:
`inotify.go` needs a tiny patch in NewWatcher.

```
  // Set IN_CLOEXEC by default. The majority of use cases should not
  // pass this descriptor down.
  fd, errno := syscall.InotifyInit1(syscall.IN_CLOEXEC)
```

nathany:
> IN_CLOEXEC
> Set the close-on-exec (FD_CLOEXEC) flag on the new file descriptor. See the description of the O_CLOEXEC flag...

> O_CLOEXEC (Since Linux 2.6.23)
> Enable the close-on-exec flag for the new file descriptor. Specifying this flag permits a program to avoid additional fcntl(2) F_SETFD operations to set the FD_CLOEXEC flag. Additionally, use of this flag is essential in some multithreaded programs since using a separate fcntl(2) F_SETFD operation to set the FD_CLOEXEC flag does not suffice to avoid race conditions where one thread opens a file descriptor at the same time as another thread does a fork(2) plus execve(2).

@pattyshack is working on a pull request.

What about support on ARM and PPC64? /cc @suihkulokki @TheTincho @clnperez @laboger @tophj-ibm?

TheTincho:
@nathany You mean support for inotify_init1? I can test building this on those arches if it is just that.

nathany:
@TheTincho That would be very helpful. inotify_init1 and the flags it takes.

I wonder if we should be using `IN_NONBLOCK` as well? https://github.com/fsnotify/fsnotify/blob/master/inotify_poller.go#L48

nathany:
Is there any reason why someone would want the child process to have the inotify file descriptor carried over? 

Presumably a new one would be created in the child process instead if needed.

pattyshack:
If you're share the inotify fd between between processes, the event
notifications may end up in either process.  In general, this behavior does
not seem particularly useful.

kqueue behaves the same way (i.e., child do not inherit kqueue) for the
same reason.

TheTincho:
I have just tried PPC64 and ARMel, and it compiles and passes the tests correctly with this patch applied. Do you need me to test somewhere else too?

PR #178, opened 2016-10-05
Create inotify fd with close-on-exec

pattyshack:
resolves #155

(originally from #174)

PR #219, opened 2017-09-01
Also create epoll and pipe fds with close-on-exec (#155)

JohannesEbke:
* Add the unix.O_CLOEXEC to the Pipe2 call
* Add unix.EPOLL_CLOEXEC to the Epoll call

I've signed the CLA.

When creating a fsnotify.NewWatcher(), four file descriptors are opened: One for inotify, one for epoll and two for the two ends of a pipe. Only the inotify one has CLOEXEC set. This PR adds the CLOEXEC flag to all four file descriptors, preventing a leak in cases where execve is used e.g. to reload an executable on code change.

I read the manpages of pipe2 and epoll_create2 to get an idea of the flags.

In a simple program which just opens a NewWatcher, does a syscall.Exec, and sleeps, the difference in the number of open file descriptors is visible in /proc/<pid>/fd/

### DIFF
--- a/inotify_poller.go
+++ b/inotify_poller.go
@@ -40,12 +40,12 @@ func newFdPoller(fd int) (*fdPoller, error) {
 	poller.fd = fd
 
 	// Create epoll fd
-	poller.epfd, errno = unix.EpollCreate1(0)
+	poller.epfd, errno = unix.EpollCreate1(unix.EPOLL_CLOEXEC)
 	if poller.epfd == -1 {
 		return nil, errno
 	}
 	// Create pipe; pipe[0] is the read end, pipe[1] the write end.
-	errno = unix.Pipe2(poller.pipe[:], unix.O_NONBLOCK)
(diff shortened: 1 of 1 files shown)


pwaller:
I can confirm that this fixes a problem I see, too. Bit surprised this wasn't fixed in #178.

sipsma:
Is there anything preventing this fix from being merged (other than needing a re-base on master)? This PR is from 2017 but the issue of these FDs leaking across execs is still present today.

JohannesEbke:
I've rebased the patch onto current master, everything still seems to work fine. It would be good to have it merged.

cpuguy83:
I just force-pushed a no-op change to try and re-trigger the CI.

cpuguy83:
ugh 😞 CI issues due to golint not working on go1.8 anymore.
Also appveyor/branch is not doing anything need to look into that as well. Can't merge until these are cleared up.

cpuguy83:
Finally, all green. Merging. Thanks! 🎉 ❤️

pwaller:
:tada:

PR #225, opened 2017-11-27
Return error from watcher Close()

cosql:
Return error from watcher Close()

Return error from Watcher.Close() in kqueue.go

Close() in kqueue.go

Check if error can be return from Close(), it is always return `nil` for now.


nhooyr:
Never mind, this should be closed. Turns out returning a nil error is the expected behaviour and there is a test for it.

nhooyr:
Definitely odd how the error is kept track of though, I removed that unnecessary code in #230

PR #230, opened 2017-12-29
Fix possible deadlock on closing the watcher on kqueue

### LINKED ISSUE #187: osx kqueue.go deadlocks on Close
OSX:
ProductName:    Mac OS X
ProductVersion: 10.11.6
BuildVersion:   15G31

closing a watcher can result in deadlock. Reviewing the kqueue.go code, there are several "raw sends" that are not protected by a select that includes the done channel. Will have PR shortly to fix this. Compare kqueue.go to the inotify.go/linux code, where all sends are protected.

nhooyr:
if the watcher is closed from within the goroutine listening on the events or errors channel, then no one will be receiving on the events or errors channel which may lead the readEvents goroutine to block on a send to one of those channels. Then the close call on the watcher will never return because the readEvents goroutine will never receive from the done channel leading the close call to block.

This PR fixes that by using a select statement in readEvents every time it wants to send on a channel to ensure that it will also always return if the watcher is closed without a deadlock.

Thanks to @glycerine for working on this in #188.

This PR closes #188 #187 #145 #211 #225


nhooyr:
Already filed out the CLA.

PR #233, opened 2018-01-17
kqueue: Make watcher.Close() O(n) instead of O(n^2)

nicks:

Fixes a performance problem in the kqueue-based implementation of watcher.

In the old implementation, watcher.Close() would clone the list of watches, then run Remove. Each run of Remove would also iterate over the full list of watches.

This indexes the watches better so that Remove doesn't need to iterate over the full list.

There should be no functional changes in this PR

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -27,13 +27,14 @@ type Watcher struct {
 
 	kq int // File descriptor (as returned by the kqueue() syscall).
 
-	mu              sync.Mutex        // Protects access to watcher data
-	watches         map[string]int    // Map of watched file descriptors (key: path).
-	externalWatches map[string]bool   // Map of watches added by user of the library.
-	dirFlags        map[string]uint32 // Map of watched directories to fflags used in kqueue.
-	paths           map[int]pathInfo  // Map file descriptors to path names for processing kqueue events.
-	fileExists      map[string]bool   // Keep track of if we know this file exists (to stop duplicate create events).
-	isClosed        bool              // Set to true when Close() is first called
+	mu              sync.Mutex                  // Protects access to watcher data
+	watches         map[string]int              // Map of watched file descriptors (key: path).
+	watchesByDir    map[string]map[int]struct{} // Map of watched file descriptors indexed by the parent directory (key: dirname(path)).
+	externalWatches map[string]bool             // Map of watches added by user of the library.
+	dirFlags        map[string]uint32           // Map of watched directories to fflags used in kqueue.
+	paths           map[int]pathInfo            // Map file descriptors to path names for processing kqueue events.
+	fileExists      map[string]bool             // Keep track of if we know this file exists (to stop duplicate create events).
+	isClosed        bool                        // Set to true when Close() is first called
 }
 
 type pathInfo struct {
@@ -51,6 +52,7 @@ func NewWatcher() (*Watcher, error) {
 	w := &Watcher{
 		kq:              kq,
 		watches:         make(map[string]int),
+		watchesByDir:    make(map[string]map[int]struct{}),
 		dirFlags:        make(map[string]uint32),
 		paths:           make(map[int]pathInfo),
 		fileExists:      make(map[string]bool),
@@ -120,6 +122,13 @@ func (w *Watcher) Remove(name string) error {
 	w.mu.Lock()
 	isDir := w.paths[watchfd].isDir
 	delete(w.watches, name)
+
+	parentName := filepath.Dir(name)
+	delete(w.watchesByDir[parentName], watchfd)
+	if len(w.watchesByDir[parentName]) == 0 {
+		delete(w.watchesByDir, parentName)
+	}
+
 	delete(w.paths, watchfd)
 	delete(w.dirFlags, name)
 	w.mu.Unlock()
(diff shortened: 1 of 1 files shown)


kqueue.go
```diff
@@ -119,6 +121,16 @@ func (w *Watcher) Remove(name string) error {
 	w.mu.Lock()
 	isDir := w.paths[watchfd].isDir
 	delete(w.watches, name)
+
+	parentName := filepath.Clean(filepath.Dir(name))
```
nhooyr: I believe the filepath.Clean() is unneeded. From my look at https://golang.org/src/path/filepath/path.go?s=13376:13404#L452 it seems Dir() already cleans it.

kqueue.go
```diff
@@ -119,6 +121,16 @@ func (w *Watcher) Remove(name string) error {
...
+	parentWatches := w.watchesByDir[parentName]
+	for i, fd := range parentWatches {
+		if fd == watchfd {
+			w.watchesByDir[parentName] = append(parentWatches[:i], parentWatches[i+1:]...)
```
nhooyr: why don't we turn parentWatches into a `map[int]struct{}` or a `map[string]int`? then it'll be a lot easier to remove a child watch.

kqueue.go
```diff
@@ -231,7 +241,9 @@ func (w *Watcher) addWatch(name string, flags uint32) (string, error) {
 
 	if !alreadyWatching {
 		w.mu.Lock()
+		parentName := filepath.Clean(filepath.Dir(name))
```
nhooyr: no need for clean()

kqueue.go
nicks: you're right! fixed

kqueue.go
nicks: done

kqueue.go
nicks: done

kqueue.go
```diff
@@ -119,6 +121,13 @@ func (w *Watcher) Remove(name string) error {
 	w.mu.Lock()
 	isDir := w.paths[watchfd].isDir
 	delete(w.watches, name)
+
+	parentName := filepath.Dir(name)
+	delete(w.watchesByDir[parentName], watchfd)
+	if len(w.watchesByDir) == 0 {
```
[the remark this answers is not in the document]
nicks: it depends on what you mean by "need"? i think it's a nice invariant that Remove inverts the state of Add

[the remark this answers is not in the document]
nhooyr: I don't understand. I don't see how we need the if statement, it doesn't seem to be really checking for anything relevant. If the length of watchersByDir is 0, then there is nothing to delete?

kqueue.go
```diff
@@ -119,6 +121,13 @@ func (w *Watcher) Remove(name string) error {
 	w.mu.Lock()
 	isDir := w.paths[watchfd].isDir
 	delete(w.watches, name)
+
+	parentName := filepath.Dir(name)
+	delete(w.watchesByDir[parentName], watchfd)
+	if len(w.watchesByDir[parentName]) == 0 {
```
nhooyr: Feel free to add more commits instead of amending so its easier for me to tell whats changed between reviews. We can squash at the end.

kqueue.go
```diff
@@ -119,6 +121,13 @@ func (w *Watcher) Remove(name string) error {
...
+		delete(w.watchesByDir, parentName)
```
nhooyr: Still not quite understanding the point of this. If the parentName has no child watches, why do we want to delete it from watchesByDir?

nhooyr:
does this PR work if you try and watch `/`?

kqueue.go
nicks: _shrug_, i don't mind removing it.

nicks:
afaict it should still work fine if you watch "/"? filepath.Dir("/") == "/". So it will add an entry to the map on Add, and remove one on Remove.

kqueue.go
nhooyr: Ah I finally got the purpose of this. We also need to check if the watchesByDir is empty for the current name and if so, to delete it there.

kqueue.go
```diff
@@ -231,7 +235,16 @@ func (w *Watcher) addWatch(name string, flags uint32) (string, error) {
 
 	if !alreadyWatching {
 		w.mu.Lock()
+		parentName := filepath.Dir(name)
 		w.watches[name] = watchfd
+
+		watchesByDir, ok := w.watchesByDir[parentName]
+		if !ok {
+			watchesByDir = make(map[int]struct{})
```
nhooyr: we can set the initial size of the map to 1 because we know we're going to insert one element.

kqueue.go
nicks: ok, i'll add it back?

kqueue.go
nicks: oh, good point!

kqueue.go
```diff
@@ -231,7 +239,16 @@ func (w *Watcher) addWatch(name string, flags uint32) (string, error) {
 
 	if !alreadyWatching {
 		w.mu.Lock()
+		parentName := filepath.Dir(name)
 		w.watches[name] = watchfd
+
+		watchesByDir, ok := w.watchesByDir[parentName]
```
nhooyr: we don't need to do anything of this stuff if `name` ends in the file separator as that means it is the root directory and the root directory as no parent. Likewise a similar thing should happen in Remove().

nhooyr: You know what, I just realized thats how it works. If you type `cd ..` at the root of the FS, you remain in the root so in a certain sense, the root is a parent of itself. 🤔

nhooyr:
@nicks thanks so much for this change <3 😍

kqueue.go
nhooyr: So lets just leave it the way it is.

nicks:
Two years may have passed, but we're still using this code and interested in this getting merged! Is there anything I need to do to get it approved?

nathany:
* Has this been rebased on master?
* Can someone test this out on macOS/BSD?

nicks:
yep, it's rebased. we've been using this in deployed binaries for a while now with no issues. if there are tests and/or benchmarks you'd like me to add that would be helpful, happy to add them.

theckman (dismissed):
Looks to work on macOS 10.15.4. I see there are some open comments, but it seems like they are largely resolved. :+1: from me.

nicks:
rebased on latest master!

arp242 (approved):

PR #234, opened 2018-01-22
fix race condition in a integration test

nhooyr:
FatalF must not be called from outside the testing goroutine or this leads to a race condition within the testing package.

See https://golang.org/pkg/testing/#T.FailNow

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -444,7 +444,7 @@ func TestFsnotifyDeleteWatchedDir(t *testing.T) {
 	// Receive errors on the error channel on a separate goroutine
 	go func() {
 		for err := range watcher.Errors {
-			t.Fatalf("error received: %s", err)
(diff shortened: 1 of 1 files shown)


nhooyr:
Something is wrong with CI hmm.

integration_test.go
```diff
@@ -444,7 +444,7 @@ func TestFsnotifyDeleteWatchedDir(t *testing.T) {
 	// Receive errors on the error channel on a separate goroutine
 	go func() {
 		for err := range watcher.Errors {
-			t.Fatalf("error received: %s", err)
+			t.Error("error received: %s", err)
```
gdey: This should be `t.Errorf("error received: %s",err)`

nathany:
@nhooyr thanks for bringing this to our attention. #266 provides the fix.

Issue #229, opened 2017-12-28
how is the listening goroutine supposed to exit in the example?


nhooyr:
https://godoc.org/github.com/fsnotify/fsnotify#example-NewWatcher

```go
watcher, err := fsnotify.NewWatcher()
if err != nil {
    log.Fatal(err)
}
defer watcher.Close()

done := make(chan bool)
go func() {
    for {
        select {
        case event := <-watcher.Events:
            log.Println("event:", event)
            if event.Op&fsnotify.Write == fsnotify.Write {
                log.Println("modified file:", event.Name)
            }
        case err := <-watcher.Errors:
            log.Println("error:", err)
        }
    }
}()

err = watcher.Add("/tmp/foo")
if err != nil {
    log.Fatal(err)
}
<-done
```

If say the watcher closes, the listening goroutine would not close, it would just remain open forever printing zero valued errors/events.

What is the correct way to use the API to ensure the goroutine exits when the watcher is closed?

alexeykazakov:
You can use `event, ok  <-watcher.Events` to check if the channel is closed:
```
go func() {
    for {
        select {
        case event, ok := <-watcher.Events:
            log.Println("event:", event)
            if event.Op&fsnotify.Write == fsnotify.Write {
                log.Println("modified file:", event.Name)
            }
            if !ok {
                return
            }
        case err, ok := <-watcher.Errors:
            log.Println("error:", err)
            if !ok {
                return
            }
        }
    }
}()
```

nhooyr:
Notwithstanding that solution, I think the API should still be changed because its too easy to misuse and not clear.

nathany:
The API probably should be changed (see also #307) at some point, but for now can we improve the example or documentation?

nhooyr:
Going to close as it has been a long time and I'm not sure if this is relevant anymore.

PR #244, opened 2018-05-01
Check if channels are closed in the example

alexeykazakov:
This PR is just about adding a check for closed channels to the example https://github.com/fsnotify/fsnotify/blob/master/example_test.go

Without that check the goroutine will keep printing empty errors (nils) & events after the watcher is closed.

See https://github.com/fsnotify/fsnotify/issues/229

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -24,18 +24,18 @@ func ExampleNewWatcher() {
 		for {
 			select {
 			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
 				log.Println("event:", event)
 				if event.Op&fsnotify.Write == fsnotify.Write {
 					log.Println("modified file:", event.Name)
 				}
-				if !ok {
-					return
-				}
 			case err, ok := <-watcher.Errors:
-				log.Println("error:", err)
 				if !ok {
 					return
 				}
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -23,13 +23,19 @@ func ExampleNewWatcher() {
...
 		for {
 			select {
-			case event := <-watcher.Events:
+			case event, ok := <-watcher.Events:
 				log.Println("event:", event)
 				if event.Op&fsnotify.Write == fsnotify.Write {
 					log.Println("modified file:", event.Name)
 				}
-			case err := <-watcher.Errors:
+				if !ok {
```
omeid: Check the `ok` before reading from the `event`. Essentially, this needs to happen before line 27.

example_test.go
```diff
@@ -23,13 +23,19 @@ func ExampleNewWatcher() {
...
+					return
+				}
+			case err, ok := <-watcher.Errors:
 				log.Println("error:", err)
+				if !ok {
```
omeid: Same as above, check for close before reading the error. :)

example_test.go
alexeykazakov: Right :)
Fixed.

Issue #235, opened 2018-01-26
CLOSE_WRITE inotify event


valar999:
When file is changed, every write() call generate WRITE event. That is not what somebody usually want when waiting to file change. WRITE event can occur in partially written file, you get several write events, etc.

CLOSE_WRITE on the other hand occur when file actually changed (at least written) and close.
I think for inotify it is better to use CLOSE_WRITE instead of WRITE or make separate event for it.

jaymell:
I agree that it would be extremely useful to have CLOSE_WRITE available as a separate event type. Depending on WRITE events requires a lot of additional work to ensure that we aren't processing a file multiple times, or processing it before it's completely written.

radu-munteanu:
This issue was discussed a few times.

For GNU/Linux there are already a few forks doing just that (you can check mine if you like). The thing is this library was made to be cross-platform, you can't just add one new event if it has no use on all platforms.

britishben:
There's lots of related issues, so I think the devs are aware.

This is the one issue that stops me from dropping the old _inotify_ libraries for _fsnotify_, so I'm watching this with interest.

jaymell:
I totally understand the desire to add platform-specific features to a multi-platform library. But I also wonder, having no familiarity with the fsnotify code, if it's reasonable in certain situations to have event types that would just never be fired on certain platforms. We can all point to plenty of examples in various software libraries where common APIs nonetheless have platform-specific considerations associated with them.

radu-munteanu:
#191 #22 #173 #202

matthias-stone:
This is a duplicate of #191.

pablodz:
please add close_write, it's extremelly useful

pablodz:
I built an alternative https://github.com/pablodz/inotifywaitgo

PR #252, opened 2018-06-23
ADDED: CloseWrite event support

fkocik:
Hello guys,

When using **fsnotify** to be notified after big file writing under Linux, it is difficult to determine if writing process is terminated because *inotify* send `WRITE` event on each `write` system call.

The goal of this patch is to introduce a new sort of event named **CloseWrite** that match `IN_CLOSE_WRITE` semantic (https://www.systutorials.com/docs/linux/man/7-inotify/) and can be usefull when reading big files that must be completely written before access (such as zip files).

As a Linux addict, I don't know how to implement **Windows** and **OS-X** ports of this patch so I just added **CloseWrite** operation in `fsnotify.go` and update `inotify.go` for **Linux** implementation. 

So, the new operation is does not work on other OSes for now.

Under **Linux** using `fsnotify.CloseWrite` as event operation mask allow to be notified only once on big file writing, when write action is finished.

Regards.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -29,6 +29,7 @@ const (
 	Remove
 	Rename
 	Chmod
+	CloseWrite
 )
 
 func (op Op) String() string {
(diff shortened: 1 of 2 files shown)


radu-munteanu:
This pull won't get through as it's not cross-platform. This has been already discussed a few times.

Please check the references below:
#235 #202 #191 #173 #22

nhooyr:
Thank you for the PR.

Unfortunately, without cross platform support, it cannot get in.

yulibaozi:
very good, That's exactly what I want.

Issue #194, opened 2017-01-22
Missing Remove Event  when file handle open before for Linux


lbgogo88:
Before reporting an issue, please ensure you are using the latest release of fsnotify.

### Which operating system (GOOS) and version are you using?

Linux: lsb_release -a
macOS: sw_vers
Windows: systeminfo | findstr /B /C:OS

Linux:
LSB Version:	:core-4.0-amd64:core-4.0-ia32:core-4.0-noarch:graphics-4.0-amd64:graphics-4.0-ia32:graphics-4.0-noarch:printing-4.0-amd64:printing-4.0-ia32:printing-4.0-noarch
Distributor ID:	CentOS
Description:	CentOS release 5.8 (Final)
Release:	5.8
Codename:	Final

### Please describe the issue that occurred.
Missing Remove Event when got file handle already open before. See the sample.

And I review the fsnotify code(fsnotify/inotify.go/func newEvent) found that unix.IN_ATTRIB was catched then e.Op was set to Chmod when `rm -f $filename;echo >$filename` done.

After that the event was ignored by "func ignoreLinux" 

For this reason, I must watch the directory to walk around.

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

```

package main

import (
        "github.com/fsnotify/fsnotify"
        "os"
        "fmt"
)

func main() {
        var err error
        path := os.Args[1]
        watcher,_ := fsnotify.NewWatcher()

        watcher.Add(path)
        done := make(chan bool)

        os.Open(path)
        go func(){
                var ev fsnotify.Event
                for {
                        select {
                        case ev = <-watcher.Events:
                        case err = <-watcher.Errors:
                                fmt.Println("##ERR",err)
                        }
                        fmt.Println(ev)
                }
        }()

        <-done
}
```

vladlosev:
What is happening here is that Linux's inotify actually delivers two notifications when a file is deleted. It first sends `IN_ATTRIB` to indicate that the number of file links have been decremented, and follows up by `IN_DELETE_SELF` to indicate that the file is deleted from disk. When some process is has open handles to the file, the file is not actually deleted from disk until all open handles are closed, and thus the second notification is also delayed. But fsnotify propagates only the second notification and [drops](https://github.com/fsnotify/fsnotify/blob/fd9ec7deca8bf46ecd2a795baaacf2b3a9be1197/inotify.go#L310) the first one after statting the file name and seeing it gone.

Here is a `go test` for this.
[...]

PR #260, opened 2018-08-16
Do not suppress Chmod on non-existent file

kolyshkin:
Currently fsnorify suppresses a Chmod event if the file does not exist
when the event is received.

This makes it impossible to use fsnotify to detect when an opened
file is removed. In such case the Linux kernel sends IN_ATTRIB event,
as described in inotify(7) man page:

> IN_ATTRIB (*)
>        Metadata  changed—for example, permissions (e.g., chmod(2)),
>        timestamps (e.g., utimensat(2)), extended attributes  (setx‐
>        attr(2)), link count (since Linux 2.6.25; e.g., for the tar‐
>        get of link(2) and for unlink(2)), and user/group ID  (e.g.,
>        chown(2)).

(in this very case it's link count that changes).

To fix:
 * Modify the code to only suppress MODIFY and CREATE events.
 * Add a test case to verify Chmod event is delivered.

While at it, fix the comment in ignoreLinux() to use the up-to-date
terminology (event ops).

This is heavily based on https://github.com/fsnotify/fsnotify/pull/205; kudos to @vladlosev for his work.

Fixes https://github.com/fsnotify/fsnotify/issues/194.

**Update**
> Please indicate that you have signed the CLA in your pull request.

Yes I did

### DIFF
--- a/inotify_test.go
+++ b/inotify_test.go
@@ -499,3 +499,52 @@ func TestInotifyWatchList(t *testing.T) {
 		}
 	}
 }
+
+func TestInotifyDeleteOpenedFile(t *testing.T) {
+	testDir := tempMkdir(t)
+	defer os.RemoveAll(testDir)
+
+	testFile := filepath.Join(testDir, "testfile")
+
+	// create and open a file
+	fd, err := os.Create(testFile)
+	if err != nil {
+		t.Fatalf("Create failed: %v", err)
+	}
+	defer fd.Close()
+
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("Failed to create watcher: %v", err)
+	}
+	defer w.Close()
+
+	err = w.Add(testFile)
+	if err != nil {
+		t.Fatalf("Failed to add watch for %s: %v", testFile, err)
+	}
+
+	checkEvent := func(exp Op) {
+		select {
+		case event := <-w.Events:
+			t.Logf("Event received: %s", event.Op)
+			if event.Op != exp {
+				t.Fatalf("Event expected: %s, got: %s", exp, event.Op)
+			}
+		case <-time.After(100 * time.Millisecond):
+			t.Fatalf("Expected %s event not received", exp)
+		}
+	}
+
+	// Remove the (opened) file, check Chmod event (notifying
+	// about file link count change) is received
+	err = os.Remove(testFile)
+	if err != nil {
+		t.Fatalf("Failed to remove file: %s", err)
+	}
+	checkEvent(Chmod)
+
+	// Close the file, check Remove event is received
+	fd.Close()
+	checkEvent(Remove)
+}
--- a/inotify.go
+++ b/inotify.go
@@ -317,12 +317,12 @@ func (e *Event) ignoreLinux(mask uint32) bool {
 		return true
 	}
 
-	// If the event is not a DELETE or RENAME, the file must exist.
-	// Otherwise the event is ignored.
-	// *Note*: this was put in place because it was seen that a MODIFY
-	// event was sent after the DELETE. This ignores that MODIFY and
-	// assumes a DELETE will come or has come if the file doesn't exist.
-	if !(e.Op&Remove == Remove || e.Op&Rename == Rename) {
+	// If the event is Create or Write, the file must exist, or the
+	// event will be suppressed.
+	// *Note*: this was put in place because it was seen that a Write
+	// event was sent after the Remove. This ignores the Write and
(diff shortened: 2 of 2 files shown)


kolyshkin:
@tsachiherman @zellyn @nathany @nhooyr @markbates PTAL

kolyshkin:
@nathany @markbates @ppknap PTAL

kolyshkin:
The PR description and the commit message slightly updated for clarity.

kolyshkin:
It's a pity that no one took a look at this :( the issue is still valid

inotify_test.go
```diff
@@ -453,3 +453,52 @@ func TestInotifyOverflow(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Create failed: %v", err)
+	}
+	defer fd.Close()
+
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("Failed to create watcher: %v", err)
+	}
+	defer w.Close()
```
vladlosev: I would make sure to that the watcher has been closed, to be consistent with other tests:
```go
defer func() {
  w.Close()
  // Wait for the close to complete.
  time.Sleep(50 * time.Millisecond)
  isWatcherReallyClosed(t, w)
}()
```

kolyshkin: 1. This check is present in test cases which test that the watcher is closed; here's the list of such tests:
* `TestInotifyCloseRightAway`
* `TestInotifyCloseSlightlyLater`
* `TestInotifyCloseSlightlyLaterWithWatch`
* `TestInotifyCloseAfterRead`

In other test cases, including this one, we're not testing this particular aspect, so we don't care if the watcher is closed.

2. Calling `t.Fatalf()` (which `isWatcherReallyClosed()` might do) from a goroutine won't have the desired effect (as `t.Fatalf()` should be called from the same goroutine as the test), so adding this would be useless.

kolyshkin:
Rebased; addressed @vladlosev review comment.

@nathany PTAL

kolyshkin:
CI failures look legit but unrelated

kolyshkin:
@nathany PTAL

amitsingh21:
@kolyshkin could you please rebase with master, above CI failure should be fixed with this https://github.com/fsnotify/fsnotify/commit/7f4cf4dd2b522a984eaca51d1ccee54101d3414a 
waiting to get this merged to master, as we need this fix

alisson276:
I confirm that small change is working. I'm having this problem using lumberjack, because it keeps the file opened. After this change the following code starts to work:
```go
package main

import (
        "os"
	"log"

        "gopkg.in/natefinch/lumberjack.v2"
)

func main() {
	myFile := "/var/log/x.log"
	file := &lumberjack.Logger{
		Filename: myFile,
	}
        file.Write([]byte{})
        file.Close()
	if err != nil {
		log.Println(err)
	}
	defer watcher.Close()

	done := make(chan bool)
	go func() {
		for {
			select {
                                log.Println("ok:", ok)
				if !ok {
					return
				}
				log.Println("event:", event)
                                if event.Op&fsnotify.Chmod == fsnotify.Chmod {
					file.Close()
				} else if event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Remove == fsnotify.Remove {
					file.Write([]byte{})
					if err := watcher.Add(event.Name); err != nil {
						log.Println("Unable to watch file: ", event.Name, " ", err)
						os.Exit(1)
					}
					log.Println("started watching file: ", event.Name)
				}
                                log.Println("err ok:", ok)
				if !ok {
					return
				}
				log.Println("error:", err)
			}
		}
	}()

	err = watcher.Add(myFile)
	if err != nil {
		log.Println(err)
	}
	<-done
}
```

Before this change, after doing a `rm /var/log/x.log` didn't recreate the file.

kolyshkin:
Rebased

nxadm:
I validated this PR for https://github.com/nxadm/tail (a tail Go library, using fsnotify). It looks good to me and it fixes bugs related to reopening deleted files. Thanks for your work, @kolyshkin!

It would be great if this PR could be reviewed and merged.

ShansOwn:
Looks like the CI problem was resolved in #378 and #381.
Can this one be moved forward?

arp242 (approved):
Rebased on main, and seems good.

One possible concern is that CHMOD events may arrive after the REMOVE, which is what this check addresses in the first place. In this case, the cure seems worse than the disease, but should probably make a note of this in the changelog for the next version.

PR #262, opened 2018-08-22
kqueue.go: Remove timeout from reading kevents.

### LINKED ISSUE #237: High CPU usage
Hello all,
I have a file, monitored by fsnotify, in used from another process and I have a leak of CPU (attached pprof diagram). What do you think about it?
capture

paulquerna-okta:

Instead of having a timeout that would trigger the readEvents to check it's
done channel, run a seaprate go-routine that watches w.done, and directly
closes the KQueue FD.

This change allows a go-process to fully sleep and go into CPU idle mode,
instead of waking up every 100ms on macOS.

Fixes #237

### DIFF
--- a/kqueue.go
+++ b/kqueue.go
@@ -25,7 +25,7 @@ type Watcher struct {
 	Errors chan error
 	done   chan struct{} // Channel for sending a "quit message" to the reader goroutine
 
-	kq int // File descriptor (as returned by the kqueue() syscall).
+	_kq int // File descriptor (as returned by the kqueue() syscall).
 
 	mu              sync.Mutex                  // Protects access to watcher data
 	watches         map[string]int              // Map of watched file descriptors (key: path).
@@ -50,7 +50,7 @@ func NewWatcher() (*Watcher, error) {
 	}
 
 	w := &Watcher{
-		kq:              kq,
+		_kq:             kq,
 		watches:         make(map[string]int),
 		watchesByDir:    make(map[string]map[int]struct{}),
 		dirFlags:        make(map[string]uint32),
@@ -62,6 +62,7 @@ func NewWatcher() (*Watcher, error) {
 		done:            make(chan struct{}),
 	}
 
+	go w.closeKq()
 	go w.readEvents()
 	return w, nil
 }
@@ -113,7 +114,7 @@ func (w *Watcher) Remove(name string) error {
 	}
 
 	const registerRemove = unix.EV_DELETE
-	if err := register(w.kq, []int{watchfd}, registerRemove, 0); err != nil {
+	if err := register(w.kq(), []int{watchfd}, registerRemove, 0); err != nil {
 		return err
 	}
 
@@ -171,9 +172,6 @@ func (w *Watcher) WatchList() []string {
 
 // Watch all events (except NOTE_EXTEND, NOTE_LINK, NOTE_REVOKE)
 const noteAllEvents = unix.NOTE_DELETE | unix.NOTE_WRITE | unix.NOTE_ATTRIB | unix.NOTE_RENAME
-
-// keventWaitTime to block on each read from kevent
-var keventWaitTime = durationToTimespec(100 * time.Millisecond)
 
 // addWatch adds name to the watched file set.
 // The flags are interpreted as described in kevent(2).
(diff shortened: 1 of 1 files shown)


paulquerna-okta:
Closing as this was merged to master

wweir:
Expect a new release. This problem has caused me a big headache.

paulquerna-okta:
Is it possible to merge this? Just rebased it against master, and we've been running with this patch on our software for about 8 months now.

kqueue.go
```diff
@@ -256,6 +254,23 @@ func (w *Watcher) addWatch(name string, flags uint32) (string, error) {
...
 }
 
+func (w *Watcher) kq() int {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w._kq
+}
+
+func (w *Watcher) closeKq() {
+	select {
```
nathany: The select isn't really necessary if there is only a single case (no default case).

paulquerna-okta: Fixed in newest rebase.

paulquerna-okta:
Rebased against current master -- just a note we've been running this patch for 18 months to reduce CPU usage on our MacOS app.  Without it the app would consistently show up Activity Monitor's Energy tab.

dmt0:
Any chance this could be merged in?

arp242:
On FreeBSD and NetBSD this blocks forever on:

    n, err := unix.Kevent(kq, nil, events, timeout)

Called from the main event loop:

    kevents, err := read(w.kq(), eventBuffer, nil)

This was always a problem, it's just that before we didn't test these platforms in the CI.

kqueue(2) on FreeBSD says:

> If timeout is a non-NULL pointer, it specifies a maximum interval to wait for an event, which will be interpreted as a struct timespec. If timeout is a NULL pointer, kevent() waits indefinitely.

kqueue(2) on OpenBSD and macOS have the same language; I added some debug printf statements, and got:

OpenBSD:

	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...  done <nil> 2
	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...closeKq ....  done bad file descriptor -1

FreeBSD:

	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...  done <nil> 2
	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...  done <nil> 1
	read ...closeKq ....done: <nil>

So the difference is that on OpenBSD Kevent() will return if the file descriptor is closed, whereas on FreeBSD it won't. I assume it's the same on macOS, but I didn't try that yet.

I can't really find this behaviour documented anywhere, and seems to be something of an implementation detail that *may* change on macOS and OpenBSD in the future too.

Not entirely sure what the best fix for this would be, but we can't just merge this as-is due to this.

---

Some other (minor) issues:

- It seems to call unix.Close() on _kq twice? once in closeKq() and once at the end of readEvents(). You added a `unix.EBADF` error check for this, but that's only needed because it's already closed and set to -1 in closeKq().

- Also, the timeout parameters is now always nil, and `durationToTimespec()` is never used (but those are minor issues, compared to it not working at all).

---

And my printf debugging, in case it's useful to anyone else:

```
diff --git kqueue.go kqueue.go
index 1d19378..3cab633 100644
--- kqueue.go
+++ kqueue.go
@@ -304,7 +304,9 @@ func (w *Watcher) kq() int {
 func (w *Watcher) closeKq() {
        <-w.done
        w.mu.Lock()
-       unix.Close(w._kq)
+       fmt.Print("closeKq ....")
+       err := unix.Close(w._kq)
+       fmt.Println("done:", err)
        w._kq = -1
        w.mu.Unlock()
 }
@@ -405,7 +407,9 @@ loop:
        }

        // cleanup
+       fmt.Print("CLEANUP ... ")
        err := unix.Close(w.kq())
+       fmt.Println("\t", err)
        if err != nil && err != unix.EBADF {
                // only way the previous loop breaks is if w.done was closed so we need to async send to w.Errors.
                select {
@@ -560,7 +564,9 @@ func register(kq int, fds []int, flags int, fflags uint32) error {
 // read retrieves pending events, or waits until an event occurs.
 // A timeout of nil blocks indefinitely, while 0 polls the queue.
 func read(kq int, events []unix.Kevent_t, timeout *unix.Timespec) ([]unix.Kevent_t, error) {
+       fmt.Print("read ...")
        n, err := unix.Kevent(kq, nil, events, timeout)
+       fmt.Printf("  done %v %d\n", err, n)
        if err != nil {
                return nil, err
        }
```

arp242:
#124 seems to fix the same problem, and doesn't seem to have any of the above issues in a quick check. Updated branch here, but needs some more work as it's a pretty old branch: https://github.com/fsnotify/fsnotify/tree/kqueue-no-timeout

arp242:
I created #480, which should solve this issue better.

PR #271, opened 2018-10-23
Bump Go versions

HaraldNordgren:


HaraldNordgren:
Go 1.8 is failing with

```
$ go get -u golang.org/x/lint/golint
# golang.org/x/tools/go/internal/gcimporter
../../../golang.org/x/tools/go/internal/gcimporter/bexport.go:212: obj.IsAlias undefined (type *types.TypeName has no field or method IsAlias)
```
so I will drop it.

PR #280, opened 2019-01-20
Add Movedto op to manage more easily this event

pduveau:
With the current version, an os rename operation raised two operation :

    Rename with previous file name
    Create with new file name

As most of the time the application needs to have two different treatments for Rename and Create events, this PR propose to add a Moveto fsnotify new Op ORed with the Create one.

So the second event gets an op Create|Movedto on Windows and Inotify. The sequence becomes:

    Rename with previous file name
    Create|Movedto with new file name

Doing this allows application to trap this particular event without breaking compatibility with previous versions.

This PR suppress travis go 1.8.x as an internal golang package is broken and makes all PR fail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ sudo: false
 language: go
 
 go:
-  - 1.8
-  - 1.7.x
+  - 1.8.x
+  - 1.9.x
   - tip
 
 matrix:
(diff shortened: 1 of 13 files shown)


pduveau:
Hi @nathany, 
Few weeks ago, I pulled a request on fsnotify to enhance the move event. The modification is very light. I'd like to get a review of my proposal.
Best regards
Philippe

nathany:
Sorry for the delay. Why did you close this Philippe?

PR #281, opened 2019-01-25
Add CLOSE_WRITE event to monitored events

vadimpilyugin:
It adds a new event to the monitored events list.


nathany:
Sorry, but adding new events that only work on Linux and don't do anything on the other supported platforms isn't a great solution.

If you have some ideas, please discuss on #22.

PR #283, opened 2019-02-18
Added informational methods to the watcher

dgutierrez1287:
It adds a method to check if the watcher is watching a file path and one to get the current list of watched files

In inotify.go there are two methods added, CheckPath and GetWatchedFiles and there are corresponding tests in the inotify_test.go 

Add files to a watcher and then check for their existence and get the list  of watched files to compare.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -87,6 +87,27 @@ func (w *Watcher) Close() error {
 	return nil
 }
 
+// CheckPath checks if a path is being watched by the watcher or not
+func (w *Watcher) CheckPath(name string) bool {
+	name = filepath.Clean(name)
+
+	if _, ok w.watches[name]; ok {
+		return true
+	}
+	return false
+}
+
+// GetWatchedFiles returns a slice of the paths currently watched by the watcher
+func (w *Watcher) GetWatchedFiles() []string {
+	watchedFiles := make([]string, 0 , len(w.watches))
+	
+	for k := range w.watches {
+		watchedFiles = append(watchedFiles, k)
+	}
+
+	return watchedFiles
+}
+
 // Add starts watching the named file or directory (non-recursively).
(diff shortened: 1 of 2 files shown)


cpuguy83:
Since the caller is the one adding paths, couldn't this be tracked outside of the watcher?
What's the use-case for this?

dgutierrez1287:
Allowing the list to be pulled out of the watcher gives a single source of truth to answer what files are being watched and if a certain file is being watched or not. I made this addition because I am working on a backup type program that watches for file changes and allows the config (of what files are being watched) to be hot reloaded. I am using this to then compare what the watcher is watching or not watching based on the config and if a file needs to be removed or added to the watcher. It could be tracked outside of the watcher but that would lead to having two lists (the one in the watcher and the external one) and I figured since the watcher already has the information it would be better to pull it directly.

cpuguy83:
Why not attempt the add/remove and check the error status?

Checking the lists like this is racey just in general (not account for the actual memory safety issues in the current change set).

dgutierrez1287:
In my case and in most cases where a config is dictating what files are watched you still need to know what the watcher is watching in order to figure out which watches have been removed from the config.

bep:
>what the watcher is watching 

Didn't you tell the watcher what it should watch?

dgutierrez1287:
Yes you do tell the watcher what to watch but the point is to be able to not have to maintain a list outside the watcher since the watcher already has that information.

nathany:
I'm not opposed to adding something like this, but I think a feature request should probably start as a new issue with some discussion about the problem and the proposed API.

If we did implement this, we would obviously need to make it cross-platform, not just supporting inotify. I'm closing this pull request, but let's keep discussing the idea.

PR #285, opened 2019-03-07
strings.Builder instead of bytes.Buffer [Go 1.10+]

ilanpillemer:
It replaces bytes.Buffer with strings.Builder in the String() method for Op.

There is only one file changes and the only change is that bytes.Buffer is replaces with strings.Builder as since go 1.10 this is the recommended most efficient way of building a String.

Anything that prints out an Op.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -9,9 +9,9 @@
 package fsnotify
 
 import (
-	"bytes"
 	"errors"
 	"fmt"
+	"strings"
 )
 
 // Event represents a single file system notification.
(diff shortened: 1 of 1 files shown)


cpuguy83:
At the moment this project supports from go1.8 and higher.
This may need to be re-addressed, but this certainly breaks that compatibility.

ilanpillemer:
This makes perfect sense. I missed this current requirement as it is not mentioned in the documentation or contributing sections that currently there is a strict dependency on the Go 1.8 standard library API.

cpuguy83:
Could either move the string implementation to go version specific files... or make an interface for a "stringWriter" and put the implementation to make one in go version specific files.

nathany:
At this point we could probably move ahead with a change like this, since Go 1.10 has been out for a while, and people can use the older releases if necessary.

We'd need to decide if this is considered a breaking change requiring a major version number, or just a feature change.

arp242 (approved):
The minimum version was already set to Go 1.12 in v1.5.0, so the version is no longer a concern.

I rebased it on the latest master to ensure there are no conflicts and that everything still works; should be good to go now.

Issue #26, opened 2014-07-08
Improved "rename" support


aktau:
As mentioned here: https://github.com/howeyc/fsnotify/issues/104

Currently, it's not straightforward to detect a rename correctly. In most cases, one will see this:

```
RENAME file1
CREATE file2
```

Yet this doesn't guarantee that `file1` and `file2` are the same file (there could be a file creation in between, or `file1` could be moved outside a watched folder and `file2` could be moved in). To work around this, one could store the `os.FileInfo` for every watched file and use `os.SameFile` to compare it with the new file. This is a bit cumbersome though, as it requires keeping a lot of `os.FileInfo` structs in memory + bookkeeping them.

However, some platforms provide a way to atomically see what was the old name and what's the new name of a file. To quote @nathany:

> On Linux there is a "cookie" used to tie the RENAME FROM and TO events together, but this isn't currently exposed by fsnotify. I still need to do more research across each platform.

This is a tracker issue to see if there is sufficient cross-platform support to enable this in the default API.

nathany:
@jaytaylor this is still something I need to research across platforms to see what can be done. Sorry for moving so slow,

jmhodges:
Hey, this would be great to get. My justrun utility is failing for vim users because vim does Weird Things with rename on OS X.

jmhodges:
Watchdog (https://github.com/gorakhargosh/watchdog) has figured out a cross-platform API, by the way.

nathany:
@jmhodges Do you have any more details on "Weird Things" vim does? Is this when saving or is there a rename file function people are using in Vim?

nathany:
ref: https://github.com/rjeczalik/notify/issues/78

omeid:
Vim uses _[swap](http://vimdoc.sourceforge.net/htmldoc/recover.html#swap-file)_ and _[backup](http://vimdoc.sourceforge.net/htmldoc/editing.html#backup)_ files, involves rename and moving.

F21:
Was a solution ever found for this?

nathany:
not as of yet.

F21:
Do you think keeping a list of `os.FileInfo`s in a concurrently safe map like `map[string]*os.FileInfo` as mentioned by @aktau is probably the best way to get this implemented?

F21:
[...]

PR #286, opened 2019-03-08
Add Movedto op

pduveau:
With the current version, an os rename operation raised two operation :

    Rename with previous file name
    Create with new file name

As most of the time the application needs to have two different treatments for Rename and Create events, this PR propose to add a Moveto fsnotify new Op ORed with the Create one.

So the second event gets an op Create|Movedto on Windows and Inotify. The sequence becomes:

    Rename with previous file name
    Create|Movedto with new file name

Doing this allows application to trap this particular event without breaking compatibility with previous versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,29 +2,36 @@ sudo: false
 language: go
 
 go:
-  - 1.8.x
-  - 1.9.x
-  - tip
+  - "stable"
+  - "1.11.x"
+  - "1.10.x"
+  - "1.9.x"
+  - "1.8.x"
 
 matrix:
+  include:
+    - go: "stable"
+      env: GOLINT=true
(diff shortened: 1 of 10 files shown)


nathany:
Thanks. Have you looked into this for kqueue for BSD and (currently) macOS?

pduveau:
Hi @nathany, No I did not integrate this feature with kqueue because the kqueue infrastructure does not provide the same kind of event and the source file has to be refactored. I, also, don't work with and know BSD.

nathany:
Thanks. Unfortunately I can't merge this solution that would have an Event that doesn't work on BSD or (currently) macOS.

I think we need to come up with a better solution. If you have any suggestions, please mention them in #26.

PR #287, opened 2019-03-20
Update documentation for linux systems (max_user_watches)

saschagrunert:
Hey there, after debugging an issue within another project I found out that `fs.inotify.max_user_instances` also has influence on the watcher creation. I verified that with a small test tool like this:

```go
package main

import (
	"fmt"

)

func main() {
	count := 0
	for {
		_, err := fsnotify.NewWatcher()
		if err != nil || count > 999 {
			fmt.Println(err)
			break
		}
		count += 1
	}
	fmt.Printf("count == %v", count)
}
```

Reaching the 1000 watchers was not possible by adapting only `fs.inotify.max_user_watches`, but with modifying `fs.inotify.max_user_instances` it worked. I updated the documentation for that scenario.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,18 @@ Spotlight indexing on OS X can result in multiple events (see [howeyc #62][#62])
 **How many files can be watched at once?**
 
 There are OS-specific limits as to how many watches can be created:
-* Linux: /proc/sys/fs/inotify/max_user_watches contains the limit, reaching this limit results in a "no space left on device" error.
+* Linux: The following files have an influence on the overall limit:
+    * `/proc/sys/fs/inotify/max_user_watches`
+    * `/proc/sys/fs/inotify/max_user_instances`
+
+  To increase them either write the appropriate value to the file or modify
+  them permanently via `/etc/sysctl.conf`:
+  ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,7 +62,18 @@ Spotlight indexing on OS X can result in multiple events (see [howeyc #62][#62])
 **How many files can be watched at once?**
 
 There are OS-specific limits as to how many watches can be created:
-* Linux: /proc/sys/fs/inotify/max_user_watches contains the limit, reaching this limit results in a "no space left on device" error.
+* Linux: The following files have an influence on the overall limit:
```
nathany: This limit sounds like it's for the number of watchers and not the number of files? If so, I think it should be a separate question about the number of watchers.

nathany (dismissed):

README.md
saschagrunert: Its generally only about the number of watchers, so I guess we're fine here right? :)

arp242 (dismissed):

arp242 (approved):

arp242:
Cheers, thanks!

Issue #42, opened 2014-09-13
When a directory is removed, cleanup w.watches (internal)


nathany:
As reported in #41.

timshannon:
FYI, I'm seeing this same behavior in my [project](https://bitbucket.org/tshannon/freehold-sync) and can readily duplicate it.

I'm going to look at synchthing's inotify library as they seemed to have fixed it.

timshannon:
I've been able to duplicate this in a stand alone piece of code.

There is a gist here (https://gist.github.com/timshannon/603f92824c5294269797) that you can run to see this issue.

nathany:
Thanks. Is this specifically occurring on Linux?

timshannon:
No, it only happens on Windows.

Works fine on linux, i expect because linux doesn't hold onto file handles like windows.

PR #288, opened 2019-03-25
[bugfix] close handle when remWatch open in getIno

brokenjacky:
when i remWatch a directory using remWatch, and then delete the directory ,and i find the directory is still in there but i can't access because a handle is still open

see #42 ,the gist by timshannon  (https://gist.github.com/timshannon/603f92824c5294269797)

### DIFF
--- a/windows.go
+++ b/windows.go
@@ -327,6 +327,9 @@ func (w *Watcher) remWatch(pathname string) error {
 	w.mu.Lock()
 	watch := w.watches.get(ino)
 	w.mu.Unlock()
+    if e := syscall.CloseHandle(ino.handle); e != nil {
+		w.Errors <- os.NewSyscallError("CloseHandle", e)
+	}
 	if watch == nil {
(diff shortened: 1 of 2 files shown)


windows.go
```diff
@@ -298,6 +298,9 @@ func (w *Watcher) remWatch(pathname string) error {
 	w.mu.Lock()
 	watch := w.watches.get(ino)
 	w.mu.Unlock()
+    if e := syscall.CloseHandle(ino.handle); e != nil {
```
Code0x58: the leading whitespace here is spaces rather than a tab

Code0x58 (dismissed):
Looks good to me. Would this be an excuse for a `windows_test.go`?

mattn:
@brokenjacky Would you update this?

arp242:
I rebased this on the latest main and I added the test case you wrote @mattn; is that test case finished or were you still working on it?

arp242 (approved):

Issue #227, opened 2017-12-11
Add option to disable following symlinks


adriansr:
### Which operating system (GOOS) and version are you using?

GOOS: darwin
ProductName:	Mac OS X
ProductVersion:	10.12.6
BuildVersion:	16G1036

### Please describe the issue that occurred.

Under macOS, fsnotify follows symlinks on watched directories and produces notifications for changes to files residing on directories pointed by the symlink.

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

1. The usual watcher / logger:

```
package main

import (
	"log"

	"github.com/fsnotify/fsnotify"
)

func main() {
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		log.Fatal(err)
	}
	defer watcher.Close()
	err = watcher.Add("/tmp/bug")
	if err != nil {
		log.Fatal(err)
	}
	for {
		select {
		case ev := <-watcher.Events:
			log.Printf("Got event %+v", ev)
		}
	}
}
```

Operations to perform:

```
mkdir /tmp/bug
mkdir /tmp/other
ln -s /tmp/other /tmp/bug/link
touch /tmp/other/somefile
```

Events received:

```
2017/12/11 23:44:38 Got event "/tmp/bug/link": CREATE
2017/12/11 23:44:41 Got event "/private/tmp/other/somefile": CREATE
```

bep:
And this is wrong? Why?

adriansr:
@bep I understand the watching is meant not to be recursive?

If instead of linking to another directory, I just create the directory inside a watched directory, I don't get notifications for it. Why would I want symlinks to behave differently?

Also, this behavior is only displayed in macOS. Not Linux or Windows.

bep:
> @bep I understand the watching is meant not to be recursive?

I don't see any recursiveness in your example.

nathany:
see #199 and #289.

Issue #199, opened 2017-03-02
Don't follow symlinks


mprobst:
### Which operating system (GOOS) and version are you using?

LSB Version:	core-2.0-amd64:core-2.0-noarch:core-3.0-amd64:core-3.0-noarch:core-3.1-amd64:core-3.1-noarch:core-3.2-amd64:core-3.2-noarch:core-4.0-amd64:core-4.0-noarch:core-4.1-amd64:core-4.1-noarch:cxx-3.0-amd64:cxx-3.0-noarch:cxx-3.1-amd64:cxx-3.1-noarch:cxx-3.2-amd64:cxx-3.2-noarch:cxx-4.0-amd64:cxx-4.0-noarch:cxx-4.1-amd64:cxx-4.1-noarch:desktop-3.1-amd64:desktop-3.1-noarch:desktop-3.2-amd64:desktop-3.2-noarch:desktop-4.0-amd64:desktop-4.0-noarch:desktop-4.1-amd64:desktop-4.1-noarch:graphics-2.0-amd64:graphics-2.0-noarch:graphics-3.0-amd64:graphics-3.0-noarch:graphics-3.1-amd64:graphics-3.1-noarch:graphics-3.2-amd64:graphics-3.2-noarch:graphics-4.0-amd64:graphics-4.0-noarch:graphics-4.1-amd64:graphics-4.1-noarch:languages-3.2-amd64:languages-3.2-noarch:languages-4.0-amd64:languages-4.0-noarch:languages-4.1-amd64:languages-4.1-noarch:multimedia-3.2-amd64:multimedia-3.2-noarch:multimedia-4.0-amd64:multimedia-4.0-noarch:multimedia-4.1-amd64:multimedia-4.1-noarch:printing-3.2-amd64:printing-3.2-noarch:printing-4.0-amd64:printing-4.0-noarch:printing-4.1-amd64:printing-4.1-noarch:qt4-3.1-amd64:qt4-3.1-noarch:security-4.0-amd64:security-4.0-noarch:security-4.1-amd64:security-4.1-noarch
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04 LTS
Release:	14.04
Codename:	trusty

### Please describe the issue that occurred.

[...]

PR #289, opened 2019-04-06
Add AddRaw method which does not follow symlinks

Code0x58:
This is a minor† change for #199 which on the surface adds a `AddRaw` method to the public interface which does not resolve symlinks before starting a watch.
 - [x] rewrite existing `Add` methods to be called `AddRaw` and not follow symlinks
 - [x] reimplement a generic `Add` method which uses `filepath.EvalSymlinks(name)` before calling `AddRaw`
 - [x] ~~consider adding atomic symlink update method~~ (separate issue)
 - [x] document

A follow on from this would be to offer a high level interface as mentioned on my initial comment for the issue - this can be tracked in a separate issue+PR.

† this does change the behaviour of `Add` on windows as it didn't follow symlinks (which do exist) before, but now it does

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,10 +72,18 @@ func main() {
 		}
 	}()
 
+	// if this is a link, it will follow all the links and watch the file pointed to
 	err = watcher.Add("/tmp/foo")
 	if err != nil {
 		log.Fatal(err)
 	}
+
+	// this will watch the link, rather than the file it points to
+	err = watcher.AddRaw("/tmp/link")
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	<-done
 }
 ```
@@ -89,6 +97,10 @@ Please refer to [CONTRIBUTING][] before opening an issue or pull request.
 See [example_test.go](https://github.com/fsnotify/fsnotify/blob/master/example_test.go).
 
 ## FAQ
+
+**Are symlinks resolved?**
+Symlinks are implicitly resolved by [`filepath.EvalSymlinks(path)`](https://golang.org/pkg/path/filepath/#EvalSymlinks) when `watcher.Add(name)` is used. If that is not desired, you can use `watcher.AddRaw(name)` to not follow any symlinks before watching. See [example_test.go](https://github.com/fsnotify/fsnotify/blob/master/example_test.go).
+
 
 **When a file is moved to another directory is it still being watched?**
 
@@ -127,4 +139,3 @@ fsnotify requires support from underlying OS to work. The current NFS protocol d
 
 * [notify](https://github.com/rjeczalik/notify)
 * [fsevents](https://github.com/fsnotify/fsevents)
-
--- a/fen.go
+++ b/fen.go
@@ -26,8 +26,8 @@ func (w *Watcher) Close() error {
 	return nil
 }
 
-// Add starts watching the named file or directory (non-recursively).
-func (w *Watcher) Add(name string) error {
+// AddRaw starts watching the named file or directory (non-recursively). Symlinks are not implicitly resolved.
+func (w *Watcher) AddRaw(name string) error {
 	return nil
 }
 
--- a/integration_test.go
+++ b/integration_test.go
@@ -65,8 +65,8 @@ func newWatcher(t *testing.T) *Watcher {
 
 // addWatch adds a watch for a directory
 func addWatch(t *testing.T, watcher *Watcher, dir string) {
-	if err := watcher.Add(dir); err != nil {
-		t.Fatalf("watcher.Add(%q) failed: %s", dir, err)
+	if err := watcher.AddRaw(dir); err != nil {
+		t.Fatalf("watcher.AddRaw(%q) failed: %s", dir, err)
 	}
 }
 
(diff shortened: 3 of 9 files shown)


README.md
```diff
@@ -43,6 +43,9 @@ See [example_test.go](https://github.com/fsnotify/fsnotify/blob/master/example_t
 
 ## FAQ
 
+**Are symlinks followed?**
+As of v2.0.0 symlinks are no longer followed. You will have to explicitly get the underlying path with [`filepath.EvalSymlinks(path)`](https://golang.org/pkg/path/filepath/#EvalSymlinks) before passing it to the watcher if you want this behaviour.
```
nathany: Should we add an example (similar to example_test) of using this?

integration_test.go
```diff
@@ -1012,6 +1011,71 @@ func TestFsnotifyClose(t *testing.T) {
...
+		t.Fatalf("Failed to create file1: %s", err)
+	}
+	defer f1.Close()
+	if _, err := os.Create(file2); err != nil {
+		t.Fatalf("Failed to create file2: %s", err)
+	}
+
+	// symlink works for Windows too
+	if err := os.Symlink(file1, link); err != nil {
+		t.Fatalf("Failed to create sylink: %s", err)
```
nathany: typo: symlink

integration_test.go
```diff
@@ -1012,6 +1011,71 @@ func TestFsnotifyClose(t *testing.T) {
...
+	select {
+	case event := <-w.Events:
+		t.Fatalf("Event from watcher: %v", event)
+	case err := <-w.Errors:
+		t.Fatalf("Error from watcher: %v", err)
+	case <-time.After(50 * time.Millisecond):
+	}
+
+	// ~atomic link change - event
+	tmpLink := filepath.Join(testDir, "tml-link")
```
nathany: did you mean tmp-link?

integration_test.go
```diff
@@ -1012,6 +1011,71 @@ func TestFsnotifyClose(t *testing.T) {
...
+	case _ = <-w.Events:
+	case err := <-w.Errors:
+		t.Fatalf("Error from watcher: %v", err)
+	case <-time.After(50 * time.Millisecond):
+		t.Fatalf("Took too long to wait for event")
+	}
+
+}
+
+/* XXX(obristow): these would return if implemeting the high level interfaced mentioned in https://github.com/fsnotify/fsnotify/issues/199#issuecomment-480277398
```
nathany: Should we remove this test, or keep it and use EvalSymlinks?

nathany:
This would also fix #227.

README.md
Code0x58: I added a block to `example_test.go` which uses `filepath.EvalSymlinks(path)`, and linked to it here. I'm imagining having it in the standard example is better so people don't miss it.

integration_test.go
Code0x58: I pushed an amended version with these tests stripped, as I'm thinking we can dig them out of history if we want, and I don't think they'd be very meaningful if using `EvalSymlinks`.

README.md
Code0x58: Now there's both `watcher.Add(name)` and `watcher.AddRaw(name)` in the example

integration_test.go
Code0x58: I've just reinstated these tests now that `watcher.Add(name)` is back to the symlink following behaviour

integration_test.go
```diff
@@ -1007,11 +1007,74 @@ func TestFsnotifyClose(t *testing.T) {
...
+
+	err = w.AddRaw(link)
+	if err != nil {
+		t.Fatalf("Failed to add link: %s", err)
+	}
+
+	// change file 1 - no event
+	f1.Write([]byte("Hello"))
+	f1.Sync()
+	// XXX(Code0x58): doing a create here shows a CHMOD event on mac - is that an issue?
```
nathany: 🤔

Code0x58:
Would renaming to something like `AddWithoutDereferencing` be preferable?

drewwells:
Can we get a release with this? Configmaps in k8s are mounted as configmaps, so fsnotify is of no use in those situations.

ifilonenko:
What is the status on this? this would be great

fen.go
```diff
@@ -26,8 +26,8 @@ func (w *Watcher) Close() error {
 	return nil
 }
 
-// Add starts watching the named file or directory (non-recursively).
-func (w *Watcher) Add(name string) error {
+// AddRaw starts watching the named file or directory (non-recursively). Symlinks are not implicitly resolved.
+func (w *Watcher) AddRaw(name string) error {
 	return nil
```
nathany, line 31: this should probably do something

nathany:
> Would renaming to something like `AddWithoutDereferencing` be preferable?

Perhaps.

Do people want to be able to add some paths to the same watcher that follow symlinks and some that don't? If not, maybe this could be a config option instead of a new API? E.g. `NewWatcher({FollowSymlinks: true})` or `watcher.FollowSymlinks(true)`?

fen.go
Code0x58: probably - I suppose it is covered by this existing [issue](https://github.com/fsnotify/fsnotify/issues/12)

Code0x58:
one rationale for having separate adding methods in my mind is that it is explicit, so when reading a line of code, you know for sure what it will do at that point, without having to go back to the watcher's creation to see what it does, and similarly a change in one place won't have a significant behavioural change on other places. It may also be a little more convenient too, as you don't have to add `filepath.EvalSymlinks(path)` when following is off, or be stuck with following when following is on

niejian (approved):
LGTM

niejian (approved):
lgtm

nathany:
Thanks. And apologies for the delay.

We need to sort out that CI issue #373.

Code0x58:
Here's a PR for migrating away from Travis CI which shows green for master with go 1.11+ #378. I rebased a copy of this branch onto that and the build is green too [[test results](https://github.com/Code0x58/fsnotify/actions/runs/1087105544)]

nathany (dismissed):

README.md
```diff
@@ -90,6 +98,10 @@ See [example_test.go](https://github.com/fsnotify/fsnotify/blob/master/example_t
 
 ## FAQ
 
+**Are symlinks resolved?**
+Symlinks are implitily resolved by [`filepath.EvalSymlinks(path)`](https://golang.org/pkg/path/filepath/#EvalSymlinks) when `watcher.Add(name)` is used. If that is not desired, you can use `watcher.AddRaw(name)` to not follow any symlinks before watching. See [example_test.go](https://github.com/fsnotify/fsnotify/blob/master/example_test.go).
```
nathany, line 102: implitily -> implicitly

nathany:
@Code0x58 I sent you a GitHub invite. Feel free to merge this.

Would you be willing to also update the CHANGELOG, AUTHORS, and tag a new release?

Code0x58:
Thanks @nathany, I squashed the PR to avoid adding 2 other author emails for myself. It seems I can't push to master so will need an additional approval following the rebase, and might not be able to tag (not sure). Here's another PR to update the CHANGELOG and AUTHORS #380

nathany (approved):

PR #294, opened 2019-05-12
Modify `example_test.go` executable and written in main-like code.

mcauto:
I think example code is difficult and hard to understand for go beginners.

example_test.go

[here function name](https://github.com/fsnotify/fsnotify/compare/master...mcauto:master?expand=1#diff-7101a3ea121adf8d5b0acb5a2c520ca3R19) change to TestExampleNewWatcher.
```bash
$ go test -v example_test.go
```
and file change or Ctrl+C

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -16,6 +16,8 @@ import (
 	"github.com/fsnotify/fsnotify"
 )
 
+// If you want to execute example code, append prefix 'Test'.
+// ex) func TestExampleNewWatcher(t *testing.T)
 func exampleNewWatcher(t *testing.T) {
 	watcher, err := fsnotify.NewWatcher()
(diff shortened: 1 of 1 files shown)


example_test.go
```diff
@@ -8,41 +8,53 @@ package fsnotify_test
...
+	"syscall"
+	"testing"
 
 	"github.com/fsnotify/fsnotify"
 )
 
-func ExampleNewWatcher() {
+// If you want to execute example code, append prefix 'Test'.
+// ex) func TestExampleNewWatcher(t *testing.T)
+func exampleNewWatcher(t *testing.T) {
```
nathany: The point of an example is to be code that people could apply to an application. It should be using testing.T.

PR #298, opened 2019-07-16
Add inotify CLOSE events

fishman:
inotify notifies when files have finished being written to. To reliably
detect if a write to a file has finished the event necessary is
CLOSE_WRITE.

Adds CLOSE_WRITE and CLOSE_NOWRITE support for linux

On a Linux machine you can just copy a file into a folder. Close will be triggered even when a file is not changed.

Update will only trigger when a write was recently finished.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -29,6 +29,8 @@ const (
 	Remove
 	Rename
 	Chmod
+	Update
+	Close
 )
 
 func (op Op) String() string {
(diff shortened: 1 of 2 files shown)


nathany:
Sorry, but adding new events that only work on Linux and don't do anything on the other supported platforms isn't a great solution.

If you have some ideas, please discuss on #22.

PR #309, opened 2019-10-05
Add Go modules

nathany:

Add Go modules

go.mod

n/a


dylan-bourque:
One note here, it's usually a bad idea to opt in to modules without bumping the major version.

nathany:
@dylan-bourque Thanks for the heads up. Do you have any pointers as to what could break? I was thinking of this being a minor release update until you chimed in.

dylan-bourque:
It's been a while since https://gitlab.com/gifts/uuid/issues/61 so I don't recall all the details. IIRC it was a bigger problem if your repo was already v2+, but in the abstract enabling support for modules is a non-trivial change in behavior. One noticeable thing is that importers who aren't using modules will get the last commit where `go.mod` doesn't exist (so no big fixes for non-modules consumers).

In contrast, though, bumping to v2 would also mean that every importer would have to change their import path (adding `/v2` at the end) to get the update.

TBH, I don't know what the _right_ path is I just didn't want to not say anything.

nathany:
That's confusing. So rather than getting the latest commit from master, no Go-modules users would be stuck on the current 1.4.7 until they switch? (Assuming they are using Go 1.11+)?

dylan-bourque:
Maybe it's the other way around and non-modules importers always get `HEAD` and switching to modules would "jump back" to the tagged commit.

Either way, the confusion was a big part of the advice to bump major versions when enabling modules.

nathany:
I'm pinging #modules on Slack for some advice. Thanks for bringing this up.

dylan-bourque:
In was going suggest exactly that.

nathany:
> "if you are adopting modules for the first time for a pre-existing repository or set of packages that have already been tagged v2.0.0 or higher before adopting modules, then the recommended best practice is to increment the major version when first adopting modules."

It looks like https://github.com/gofrs/uuid/issues/61 was v3.

So I think we're okay.

go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/fsnotify/fsnotify
```
AlekSi, line 1: @nathany Was it an intentional change from https://gopkg.in/fsnotify.v1, or a mistake?

nathany: An intentional change.

PR #311, opened 2019-10-06
Add go.mod and go.sum

Code0x58:
These were made with:
```sh
go mod init
go get
```
This let me run commands like `go test` and `go fmt` when using `go1.13`


Code0x58:
There's already a PR for this

PR #313, opened 2019-10-16
Add Linux IN_CLOSE_WRITE support

bevand10:
Adds support for the Linux inotify IN_CLOSE_WRITE event. This is triggered when a file writer/modifier has finished making changes to a file (e.g. if use is monitoring a FTP upload folder, event is fired when a new file has finished being written).

fsnotify.go has an extra Op flag: CloseWrite
inotify.go has changes that add the IN_CLOSE_WRITE flag.

Run an app monitoring any directory. Open a file for write (e.g. using vim). On file close (:wq in vim) the IN_CLOSE_WRITE event will be triggered.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -29,6 +29,7 @@ const (
 	Remove
 	Rename
 	Chmod
+	CloseWrite
 )
 
 func (op Op) String() string {
@@ -49,6 +50,9 @@ func (op Op) String() string {
 	}
 	if op&Chmod == Chmod {
 		buffer.WriteString("|CHMOD")
+	}
+	if op&CloseWrite == CloseWrite {
+		buffer.WriteString("|CLOSEWRITE")
 	}
 	if buffer.Len() == 0 {
 		return ""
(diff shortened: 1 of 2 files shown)


simon04:
This addition has been discussed/declined already a few times in the past: #252 #235 #202 #191 #173 #22

jbrindle:
It's been 6 years and you haven't come up with "cross-platform" way to do this, could you allow options for platforms that support them and not for ones that don't, like any OS level cross platform project?

ronakg:
I'm not sure what's the issue with this particular change. It's not like it conflicts with other options in the package. `CloseWrite` provides a way to monitor another event for the users of `fsnotify`.

chamber2018:
LGTM

nathany (changes requested):
Thanks for the contribution.

`IN_CLOSE_WRITE` has come up many times over the years, but so far nobody has provided a suitable solution. `fsnotify` is a cross-platform library. Adding a `CloseWrite` op that only functions on Linux is a non-starter. Consideration must be given to how things will work on BSD, Windows, macOS, and Solaris.

nathany:
Please move this to a discussion #392.

PR #314, opened 2019-11-06
Create aix.go

dsgrangs-paul:
Empty implementation, just to allow build

It proposes an empty implementation, so that this can bu built on AIX 

Only one file to review, rather simple as all methods return either nil, or errors.New("not implemented")

This new file allows you to build the package on AIX. 
This fsnotify package is used as a dependency but the runtime is not necessary.

### DIFF
--- a/aix.go
+++ b/aix.go
@@ -1,0 +1,101 @@
+// Copyright 2019 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build aix
+
+// empty implementation, just to allow build
+
+package fsnotify
+
+import (
+	"errors"
+	"os"
+	"sync"
+	"time"
+)
+
+const (
+	sleepTime time.Duration = 50 * time.Millisecond
+)
+
+var (
+	ErrWatchedFileDeleted = errors.New("error: watched file or folder deleted")
+)
+
+// Watcher watches a set of files, delivering events to a channel.
+type Watcher struct {
+	Events    chan Event
+	Errors    chan error
+	mu        *sync.Mutex   // Map access
+	closed    chan struct{} // Channel to respond to Close
+	close     chan struct{}
+	wg        *sync.WaitGroup
+	running   bool
+	names     map[string]bool        // bool for recursive or not.
+	files     map[string]os.FileInfo // map of files.
+	ops       map[Op]struct{}        // Op filtering.
+	maxEvents int
+}
+
+// NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
+func NewWatcher() (*Watcher, error) {
+	// Set up the WaitGroup for w.Wait().
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	w := &Watcher{
+		Events: make(chan Event),
+		Errors: make(chan error),
+		closed: make(chan struct{}),
+		close:  make(chan struct{}),
+		mu:     new(sync.Mutex),
+		wg:     &wg,
+		files:  make(map[string]os.FileInfo),
+		names:  make(map[string]bool),
+	}
+	go w.readEvents()
+	return w, nil
+}
+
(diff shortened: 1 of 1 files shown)


amdprophet:
We're looking forward to using this change in https://github.com/sensu/sensu-go! Is there a timeline available for merging it?

dsgrangs-paul:
> We're looking forward to using this change in https://github.com/sensu/sensu-go! Is there a timeline available for merging it?

I wish it can be merged asap, but we need a review to be done

amdprophet:
@nathany any chance you can take a look at this soon?

amdprophet:
@tklauser do you have the power to get this merged?

tklauser:
Unfortunately I don‘t. Maybe @cpuguy83 can help getting this merged?

cpuguy83:
I don't understand why do we need the empty implementation?

echlebek:
The empty implementation is needed so that programs that depend on fsnotify can be built on AIX. The fsnotify support won't work, but in a lot of cases that's better than the program failing to build at all.

I think the empty implementation could probably be simplified. It looks like there's some unnecessary code that's not really doing anything.

cpuguy83:
If fsnotify doesn't compile on a platform, why is it imported there?

echlebek:
fsnotify in many cases is a dependency of a dependency of a dependency. If it doesn't build on AIX, then every package that imports it has to create build workarounds for AIX, should they wish to support AIX.

You might argue that it's not fsnotify's responsibility to provide this convenience. I agree, but if compatibility can be offered here, it would save rather a lot of work for rather a lot of people! :)

cpuguy83:
I'd be fine if a `fsnotify_unsupported.go` file was added if all that is needed is to keep the compiler from complaining that there's no compatible go source files, but otherwise it seems a bit off to add a dummy implementation.

Helflym:
Hi guys, I've submitted sometimes ago a PR which enables fsnotify on AIX (https://github.com/fsnotify/fsnotify/pull/326). It would be better to have this version even if not perfect than to disable fsnotify completely. 
Note that I still support that a `fsnotify_unsupported.go` should be added. This is helping a lot new GOOS coming into Go environment.

aix.go
```diff
@@ -0,0 +1,102 @@
+// Copyright 2019 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build aix
```
nathany, line 5: should add the new style build tags too

aix.go
```diff
@@ -0,0 +1,101 @@
...
+	"sync"
+	"time"
+)
+
+const (
+	sleepTime time.Duration = 50 * time.Millisecond
+)
+
+var (
+	ErrWatchedFileDeleted = errors.New("error: watched file or folder deleted")
```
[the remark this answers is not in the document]
nathany: why is this here?

aix.go
```diff
@@ -0,0 +1,102 @@
...
+		Errors: make(chan error),
+		closed: make(chan struct{}),
+		close:  make(chan struct{}),
+		mu:     new(sync.Mutex),
+		wg:     &wg,
+		files:  make(map[string]os.FileInfo),
+		names:  make(map[string]bool),
+	}
+	go w.readEvents()
+	return w, nil
```
nathany, line 59: I don't know that an empty implementation should do this

nathany: Why not just return `errors.New("Not implemented")`?

nathany (commented):
I like the idea of `fsnotify_unsupported.go` for multiple unsupported OSes, including AIX.

It might be a good idea to have some tests for this to ensure that the usage behaves appropriately when interacting with a stubbed out implementation?

tklauser:
FWIW, I've opened #424 adding a minimal `fsnotify_unsupported.go` [as suggested](https://github.com/fsnotify/fsnotify/pull/314#pullrequestreview-856173239).

PR #317, opened 2019-12-09
Changes 

dbaynak:
Added:
- Fanotify implementation based on `inotify.go` and `inotify_poller.go`;
  - added 'File' field in Event to support passing file descriptor from Fanotify event;
- support for setting a custom flag for Inotify events;
- reporting IN_ISDIR event flag in Op for Inotify.

### DIFF
--- a/fanotify.go
+++ b/fanotify.go
@@ -1,0 +1,186 @@
+package fsnotify
+
+import (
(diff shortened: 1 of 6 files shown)


dbaynak:
ignore this plz

PR #326, opened 2020-02-20
Add polling fallback

Helflym:
Polling fallback can be used by operating systems without any internal
mechanism to watch files. This is only use on AIX for now, but should
work on every over OS (tested on Linux).


Helflym:
Any news about this CL ? I need it for other go packages.

ncerny:
I need this as well.  Do we have any update on this PR?

cobolbaby:
If the number of subdirectories in the monitored directory exceeds 100k, do you need to reconsider the implementation of the function and make directory traversal as efficient as possible.

Helflym:
> @Helflym If the number of subdirectories in the monitored directory exceeds 100k, do you need to reconsider the implementation of the function and make directory traversal as efficient as possible.

This implementation was made in order to enable fsnotify on AIX, but it's clearly not optimized. I haven't made any performance metrics but I guess it won't work very well with more than 100k directories. 
If you wish to provide an optimized version, go on. I, personally, don't have time to do so for now.

Issue #21, opened 2014-06-29
Subtree watch on Windows


nathany:
@htoothrot mentioned at https://github.com/howeyc/fsnotify/issues/56#issuecomment-25312613

> ReadDirectoryChanges[1] has a "bWatchSubtree" argument. If this argument is true, you will receive events for all files/directories for the whole directory tree rooted at the given directory.
> 
> As watching a large/busy tree can lead to many events I thought this was worth bringing up: It may be helpful to have the buffer size configurable in some way. Right now it's a hard coded 4096 bytes. When many events fire quickly, events can be dropped when the buffer is full. There are caveats to simply increasing the size (http://stackoverflow.com/a/3250454 explains some of the issues) and it seems there's not one answer to fit all uses.

Also look into [FindFirstChangeNotification](http://msdn.microsoft.com/en-ca/library/windows/desktop/aa364417%28v=vs.85%29.aspx)

cup:
https://github.com/fsnotify/fsnotify/issues/18

Issue #18, opened 2014-06-29
User-space recursive watcher


nathany:
@jbowtie requested this feature at https://github.com/howeyc/fsnotify/issues/56.

Windows and FSEvents (OS X) can watch a directory tree, in fact FSEvents is built entirely for this purpose:

> if you monitor a directory with FSEvents then the monitor is recursive; there is no non-recursive option - @samjacobson https://github.com/howeyc/fsnotify/issues/54#issuecomment-24040227

Ideally fsnotify would expose a way to use this functionality on Windows and Macs.

Unfortunately inotify (Linux, Android), kqueue (BSD, iOS), and FEN (Solaris) don't have this capability.

There is some debate as to whether or not fsnotify should include a user-space recursive watcher to make up the difference. If not, perhaps it can be implemented a separate wrapper or utility library. In any case, it's something often requested, so we should definitely be thinking about it.
### Implementation

In two parts:
1. Walk subdirectories to Add (or Remove) watches for each directory:
   - To avoid watching too much, it may be important to (optionally?) skip hidden directories (.git, .hg).
   - My proposal to configure Ops filtering globally at the Watcher level (https://github.com/fsnotify/fsnotify/issues/7) simplifies recursive watching.
2. Listen for incoming Create events to watch additional directories as they are created (again, avoiding hidden directories).

purpleidea:
Is the implementation in the https://github.com/xyproto/recwatch package correct, or are there issues? I ask because it looks as if this is a more subtle, tricky problem. cc @nathany 

Cheers

nathany:
It should work. Most of the tricky issues are with too many resources being consumed. Also, we would ideally take advantage of native recursive watching on the operating systems that support it.

landaire:
Any update on this? It'd be great to have.

nathany:
No update for fsnotify, but take a look at https://github.com/rjeczalik/notify as an alternative.

xyproto:
I use [recwatch](https://github.com/xyproto/recwatch) in [algernon](https://github.com/xyproto/algernon), which works great on Linux/Windows/OS X. The only gotcha is to watch out for OS limitations in how many files can be watched. Especially on OS X the ulimit is very low by default.

nathany:
Rafal's notify library uses FSEvents on OS X instead of kqueue, so it doesn't have that problem on OS X.

xyproto:
That is nice. It has many unresolved issues, though.

nathany:
[...]

Issue #259, opened 2018-08-14
Renaming folder which already added to watch 


gohelraj1111:
### Operating system and version.

OS Name:                   Microsoft Windows 10 Pro
OS Version:                10.0.17134 N/A Build 17134
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Standalone Workstation
OS Build Type:             Multiprocessor Free

### Here is the description of the issue that occurred.

While creating new folder if I rename that folder to any custome name and than if I will create any files into that folder  at that time watcher try to find that files into old folder path which was created by system not on the given custome name

1) Output when I first create folder with custome name.
It adds both the folder path into watches
![creatrenamedir](https://user-images.githubusercontent.com/42057556/44095122-9f0edd5e-9ff5-11e8-9069-45f770a55fb2.PNG)
 
2) When I create any file into renamed folder it try to finds that file into old name folder.
![creatrenamedir2](https://user-images.githubusercontent.com/42057556/44095292-0f0782aa-9ff6-11e8-97f6-a27ebb100b08.PNG)

I also tried removing old directory path from watcher but it not removing it actually..!

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

1. Add a directory to watch
2. Create new folder in that directory by right clicking in windows explorer and give it name you like
3. When watcher fires create event then add newly created directory to watch

arnegroskurth:
We are also experiencing this issue on Windows 10.

@dylan-bourque Any update on this?

arnegroskurth:
Same issue: https://github.com/fsnotify/fsnotify/issues/243

nathany:
Thanks for reporting the issue. Have you managed to find a workaround? Is it possible to detect the rename? Possibly to remove the old watch and set up a new watch when the rename happens?

If you find a work around, it would be great to get a failing test case and fix upstreamed into fsnotify.

Though word of warning, the fsnotify code for Windows could use some love. It was pulled from winfsnotify that initially mapped Windows to Linux inotify before being further wrapped in the fsnotify common interface.

Issue #243, opened 2018-04-20
Creating a file inside a folder created using windows explorer generates events for the wrong dir


vaghetti:
Before reporting an issue, please ensure you are using the latest release of fsnotify.

### Which operating system (GOOS) and version are you using?
OS Name:                   Microsoft Windows 10 Enterprise
OS Version:                10.0.14393 N/A Build 14393
OS Manufacturer:           Microsoft Corporation
OS Configuration:          Member Workstation
OS Build Type:             Multiprocessor Free

### Please describe the issue that occurred.
Creating a file inside a folder created through windows explorer triggers an event for the `New folder` name, not the one actually chosen.

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

- Run this program:
- Create a new folder in the watched directory using windows explorer (it works if you just use powershell). Give it some name.
- Create a new folder or file or folder inside the folder you just created.
- You should see the error attached below

```
package main

import (
	"fmt"
	"log"

	"github.com/fsnotify/fsnotify"
)

func main() {
	watcher, err := fsnotify.NewWatcher()
	if err != nil {
		log.Fatal(err)
	}

	err = watcher.Add(".")
	if err != nil {
		log.Fatal(err)
	}

	for ev := range watcher.Events {
		fmt.Printf("got event %+v\n", ev)
		if ev.Op == fsnotify.Create {
			err = watcher.Add(ev.Name)
			fmt.Println("adding ", ev.Name, "to watch list")
			if err != nil {
				log.Fatal(err)
			}
		}
	}
}
```

You should see something like:
```
# using windows explorer, click on "New Folder"
got event "New folder": CREATE
adding  New folder to watch list
# pick the name abc for the folder
got event "New folder": RENAME
got event "abc": CREATE
adding  abc to watch list
# create another folder inside abc and you'll see the issue. It reports the new folder being created inside "New folder", not the chosen name "abc"
got event "New folder\\New folder": CREATE
adding  New folder\New folder to watch list
# naturally, it fails because the folder doesn't exist so it can't be watched
2018/04/20 17:26:42 GetFileAttributes: The system cannot find the path specified.
exit status 1
```

If I run the program again and create a file inside `abc`, the events are triggered as expected.

dylan-bourque:
Same underlying issues as #17.  The watch on the renamed folder is hanging onto the "old" folder name.

PR #339, opened 2020-05-07
windows: expose support for recursive directory watches

nicks:
On Windows, ReadDirectoryChanges allows recursive watches, so I added a function call to enable it

I saw some discussion on this in these issues: 
https://github.com/fsnotify/fsnotify/issues/18
https://github.com/fsnotify/fsnotify/issues/21

but given that there's not much activity in this repo, it didn't seem worth blocking on a user-space solutoin to expose this

I've run the unittest on Windows and confirms it watches things successfully. There's also a new test in integration_test, adapted from the existing subdirectory test.

### DIFF
--- a/fen.go
+++ b/fen.go
@@ -9,6 +9,7 @@ package fsnotify
 
 import (
 	"errors"
+	"fmt"
 )
 
 // Watcher watches a set of files, delivering events to a channel.
@@ -20,6 +21,13 @@ type Watcher struct {
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
 func NewWatcher() (*Watcher, error) {
 	return nil, errors.New("FEN based watcher not yet supported for fsnotify\n")
+}
+
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
+	return fmt.Errorf("Not supported")
 }
 
 // Close removes all watches and closes the events channel.
--- a/fen.go
+++ b/fen.go
@@ -9,7 +9,6 @@ package fsnotify
 
 import (
 	"errors"
-	"fmt"
 )
 
 // Watcher watches a set of files, delivering events to a channel.
@@ -27,7 +26,7 @@ func NewWatcher() (*Watcher, error) {
 // if this file system supports that.
 // Must be called before Add/Remove.
 func (w *Watcher) SetRecursive() error {
-	return fmt.Errorf("Not supported")
+	return errors.New("Not supported")
 }
 
 // Close removes all watches and closes the events channel.
--- a/inotify.go
+++ b/inotify.go
@@ -74,7 +74,7 @@ func (w *Watcher) isClosed() bool {
 // if this file system supports that.
 // Must be called before Add/Remove.
 func (w *Watcher) SetRecursive() error {
-	return fmt.Errorf("Not supported")
+	return errors.New("Not supported")
 }
 
 // Close removes all watches and closes the events channel.
--- a/inotify.go
+++ b/inotify.go
@@ -70,6 +70,13 @@ func (w *Watcher) isClosed() bool {
 	}
 }
 
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
+	return fmt.Errorf("Not supported")
+}
+
 // Close removes all watches and closes the events channel.
 func (w *Watcher) Close() error {
 	if w.isClosed() {
--- a/integration_test.go
+++ b/integration_test.go
@@ -1239,6 +1239,105 @@ func TestRemoveWithClose(t *testing.T) {
 	}
 }
 
+func TestSetRecursive(t *testing.T) {
+	watcher := newWatcher(t)
+	err := watcher.SetRecursive()
+	expectRecursive := err == nil
+	isWindows := runtime.GOOS == "windows"
+	if isWindows != expectRecursive {
+		t.Fatalf("Only expected SetRecursive to succeed on windows")
+	}
+
+	// Create directory to watch
+	testDir := tempMkdir(t)
+	defer os.RemoveAll(testDir)
+
+	testSubDir := filepath.Join(testDir, "sub")
+	testSubDirFile := filepath.Join(testDir, "sub/TestFsnotifyFile1.testfile")
+
+	// Receive errors on the error channel on a separate goroutine
+	go func() {
+		for err := range watcher.Errors {
+			t.Fatalf("error received: %s", err)
+		}
+	}()
+
+	// Receive events on the event channel on a separate goroutine
+	eventstream := watcher.Events
+	var createReceived, deleteReceived counter
+	done := make(chan bool)
+	go func() {
+		for event := range eventstream {
+			// Only count relevant events
+			expected := event.Name == filepath.Clean(testDir) || event.Name == filepath.Clean(testSubDir)
+			if expectRecursive {
+				expected = expected || event.Name == filepath.Clean(testSubDirFile)
+			}
+			if expected {
+				t.Logf("event received: %s", event)
+				if event.Op&Create == Create {
+					createReceived.increment()
+				}
+				if event.Op&Remove == Remove {
+					deleteReceived.increment()
+				}
+			} else {
+				t.Logf("unexpected event received: %s", event)
+			}
+		}
+		done <- true
+	}()
+
+	addWatch(t, watcher, testDir)
+
+	// Create sub-directory
+	if err := os.Mkdir(testSubDir, 0777); err != nil {
+		t.Fatalf("failed to create test sub-directory: %s", err)
+	}
+
+	// Create a file (Should not see this! we are not watching subdir)
+	var fs *os.File
+	fs, err = os.OpenFile(testSubDirFile, os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		t.Fatalf("creating test file failed: %s", err)
+	}
+	fs.Sync()
+	fs.Close()
+
+	time.Sleep(200 * time.Millisecond)
+
+	// Make sure receive deletes for both file and sub-directory
+	os.RemoveAll(testSubDir)
+
+	// We expect this event to be received almost immediately, but let's wait 500 ms to be sure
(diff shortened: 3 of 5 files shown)


cobolbaby:
LGTM.

@houndci-bot , could merge this PR?

@nicks  Another question is, after configuring the `recursive` option of `syscall.ReadDirectoryChanges`, will it hold a lot of memory like the recursive directory `addWatch`?

The problem I currently encounter is that due to the limitations of business scenarios, there will be a lot of monitoring subdirectories, and the recursive `addWatch` operation of a directory will occupy more than 1G of memory space, which obviously takes up a lot of resources.

hu13:
@nicksantos if this gets merged, you may need to include this https://github.com/fsnotify/fsnotify/pull/361 when your recursive watch hits a filepath that exceeds windows' syscall.MAX_PATH.

fen.go
```diff
@@ -21,6 +22,13 @@ func NewWatcher() (*Watcher, error) {
 	return nil, errors.New("FEN based watcher not yet supported for fsnotify\n")
 }
 
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
+	return fmt.Errorf("Not supported")
```
mattn, line 29: Please use errors.New

inotify.go
```diff
@@ -69,6 +69,13 @@ func (w *Watcher) isClosed() bool {
 	}
 }
 
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
+	return fmt.Errorf("Not supported")
```
mattn, line 76: ditto

integration_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestRemoveWithClose(t *testing.T) {
...
+	if isWindows != expectRecursive {
+		t.Fatalf("Only expected SetRecursive to succeed on windows")
+	}
+
+	// Create directory to watch
+	testDir := tempMkdir(t)
+	defer os.RemoveAll(testDir)
+
+	testSubDir := filepath.Join(testDir, "sub")
+	testSubDirFile := filepath.Join(testDir, "sub/TestFsnotifyFile1.testfile")
```
mattn, line 1243:
```
filepath.Join(testDir, "sub", "TestFsnotifyFile1.testfile")
```

mattn: Or
```
testSubDirFile := filepath.Join(testSubDir, "TestFsnotifyFile1.testfile")
```

integration_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestRemoveWithClose(t *testing.T) {
...
+	done := make(chan bool)
+	go func() {
+		for event := range eventstream {
+			// Only count relevant events
+			expected := event.Name == filepath.Clean(testDir) || event.Name == filepath.Clean(testSubDir)
+			if expectRecursive {
+				expected = expected || event.Name == filepath.Clean(testSubDirFile)
+			}
+			if expected {
+				t.Logf("event received: %s", event)
```
mattn, line 1264: Please use `%v` instead of `%s`

integration_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestRemoveWithClose(t *testing.T) {
...
+				if event.Op&Create == Create {
+					createReceived.increment()
+				}
+				if event.Op&Remove == Remove {
+					deleteReceived.increment()
+				}
+			} else {
+				t.Logf("unexpected event received: %s", event)
```
mattn, line 1272: ditto

integration_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestRemoveWithClose(t *testing.T) {
...
+			}
+		}
+		done <- true
+	}()
+
+	addWatch(t, watcher, testDir)
+
+	// Create sub-directory
+	if err := os.Mkdir(testSubDir, 0777); err != nil {
+		t.Fatalf("failed to create test sub-directory: %s", err)
```
mattn, line 1282: Use `%v`

integration_test.go
```diff
@@ -1226,6 +1226,105 @@ func TestRemoveWithClose(t *testing.T) {
...
+	}
+
+	// Create a file (Should not see this! we are not watching subdir)
+	var fs *os.File
+	fs, err = os.OpenFile(testSubDirFile, os.O_WRONLY|os.O_CREATE, 0666)
+	if err != nil {
+		t.Fatalf("creating test file failed: %s", err)
```
mattn, line 1289: ditto

kqueue.go
```diff
@@ -63,6 +63,13 @@ func NewWatcher() (*Watcher, error) {
 	return w, nil
 }
 
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
+	return fmt.Errorf("Not supported")
```
mattn, line 70: Use errors.New

nathany:
@mattn Thanks for doing the review. Since this PR is nearly 2 years old, if @nicks doesn't respond, it may be necessary to `gh pr checkout` this branch and open a new pull request with the suggested fixes.

nathany:
#397

nicks:
thanks! addressed all comments. most of the comments were on to changed things that were also issues in other tests, so i took the liberty of addressing those too. also rebased on main.

windows.go
```diff
@@ -37,17 +38,29 @@ func NewWatcher() (*Watcher, error) {
...
+		recursive: false,
 	}
 	go w.readEvents()
 	return w, nil
 }
 
+// SetRecursive makes future Add() calls watch directories recursively
+// if this file system supports that.
+// Must be called before Add/Remove.
+func (w *Watcher) SetRecursive() error {
```
nathany, line 56: I'm not sure about this API -- what do others think?

Adding a separate API is good in that there aren't any breaking changes for existing users.

There is no way to disable the recursive flag. Is that important?

nicks: ya, i could go either way on this.

my thinking was that you want the API to be as restrictive as possible - no switching back and forth between recursive and non-recursive, no switching modes once the watcher has started listening on events.

But certainly `SetRecursive(bool)` would be a lot more idiomatic.

nathany: There are a lot of different options, and it's important to get the API right, because the same API would need to support a user-space recursive watcher (#18) and [FSEvents](https://github.com/fsnotify/fsevents) on macOS.

@pbnjay @rogpeppe Having look at the Mac side, do you have any opinions on how this API should work?

If I recall, macOS is always recursive when making a watch?

This option seems pretty good and is backwards compatible -- a method on the watcher.

Other options would be some other form of NewWatcher with configuration. If we want the watcher to be 100% recursive or 100% not, this may be better.

Or having a separate `AddRecursive` function, which would allow intermixing of recursive and not-recursive. But we may wind up with too many variants -- like with and without Symlinks, Recursive, Ops filtering, etc.

hu13: @nathany 

I can confirm that fsevents on macOS is recursive by default.

I am in support of `Other options would be some other form of NewWatcher with configuration. If we want the watcher to be 100% recursive or 100% not, this may be better`.

I don't see a concrete use case in which we want to keep switching back and forth. Most of time, people want a recursive watch. If there are paths that they are not interested in, they can reasonably add additional filtering-out step at their application level to ignore such events.

Please let us know what else to do for this to be acceptable. thanks

hu13: As a side note, we forked this and enabled a recursive watch like this in our production software and we don't observe issues https://github.com/fsnotify/fsnotify/issues/259 and https://github.com/fsnotify/fsnotify/issues/243.

nathany: I think we could just go with this API for now. 

It's possible to change it later in a v2 of the library.

We probably just need someone to update the branch, merge it, and do the release

pbnjay: (Sorry this somehow skipped my inbox)

Yes macOS is recursive by default, but it could be a challenge to implement logic to "disable" that however. It sounds like could be the default fsnotify semantics?

arp242: One possibility would be to use `Add("dir/**")` or `Add("dir/...")`.

Especially since recursive operations may be expensive on some platforms, there may be use cases where you might want things like "recursively watch ~/.cache/go-build, and non-recursively watch ~/.cache/go-mod", or something like that.

The advantage of using a string like that is that #352 can be added without too much friction too as e.g. `AddMask("path", fsnotify.CREATE | fsnotify.DELETE)` or some such.

nathany: Would it make sense to have a syntax like Glob?

https://pkg.go.dev/path/filepath@go1.18.4#Glob
https://pkg.go.dev/io/fs@go1.18.4#Glob

nathany: One challenge with this approach is what happens on BSD (kqueue) or Linux (inotify), since only Windows and FSEvents on macOS support recursive watching. 

Is it an error on some platforms? At least until such time as a user-space recursive watcher (#18) covers those platforms?

arp242: > One challenge with this approach is what happens on BSD (kqueue) or Linux (inotify), since only Windows and FSEvents on macOS support recursive watching.

Yeah, .Add() can return an error in those cases, similar to how SetRecursive() does now. Something like:


	func recursive(path string) (string, bool) {
		if strings.HasSuffix(path, "/...") || strings.HasSuffix(path, `\...`) {
			return path[:len(path)-4], true
		}
		return path, false
	}

	path, recurse := recursive(path)
	if recurse {
    	return errors.New("recursion not supported on this system")
	}

Or for Windows and other systems where it works:

    recurse := false
    path, recurse = recursive(path)


> Would it make sense to have a syntax like Glob?

I'd be in favour of that; there's an old issue for that (#10).

Using the `/...` syntax is probably better, since e.g. something like `**/*.zip` won't work and I don't think we need to support that (it'll be tricky to get that to work well for all cases).

nathany: Yeah, that makes sense, and there is precedence for that format (e.g. `go test ./...`). Sounds good.

arp242:
There's still quite a bit of work that needs to be done in terms of API; not 100% sure what it will look like yet, but certainly different from this patch. For example #521 adds a very different way of passing options, and I probably want to use `path/...` anyway rather than options.

The core is pretty trivial (just pass a boolean around), just need to figure out an API that works well. Either way, much of this will have to be re-done, and I don't think there's much value in keeping this PR open or basing it this PR, as most things will be different.

Can subscribe to #21 for notification when it's added.

PR #349, opened 2020-11-10
Fix build on Plan 9

fhs:
This fixes build for programs that import fsnotify but fsnotify is not
absolutely necessary for the program to be useful.

Fixes build on Plan 9.

plan9.go

`GOOS=plan9 go build`


PaluMacil (approved):
seems sensible and safe to merge

arp242:
#424 made this one superfluous: it does the same for all unsupported systems (including Plan 9).

PR #350, opened 2020-11-10
TravisCI: bump Go versions

fhs:
This fixes the CI. Builds were failing because `go.mod` file wasn't
being used on older Go versions on macOS. This led to using the wrong
version of `golang.org/x/sys/unix`. See for example:
https://travis-ci.org/github/fsnotify/fsnotify/builds/742594311

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ language: go
 
 go:
   - "stable"
-  - "1.11.x"
-  - "1.10.x"
-  - "1.9.x"
+  - "1.15.x"
+  - "1.14.x"
+  - "1.13.x"
 
 matrix:
   include:


PaluMacil (approved):
Seems correct to me. I think this version bump is good

PaluMacil (dismissed):
looks good; please merge. No concerns

kolyshkin:
This needs to be updated to add 1.15 and drop 1.12.

fhs:
> This needs to be updated to add 1.15 and drop 1.12.

Done

kolyshkin:
@fhs thank you!

Frankly speaking, we need to switch away from travis (which is going to be retired real soon now, see travis-ci.org). I found that github actions is currently a good alternative.

I'd prepare a PR, but alas this repo seems abandoned (see e.g. https://github.com/fsnotify/fsnotify/pull/260).

fhs:
ping @fsnotify, @nathany

nathany:
We've switched to GitHub Actions. Thanks for the contribution

PR #359, opened 2021-02-13
add getters for events and errors so that fsnotify can be more easily mocked

daniel-garcia:
This PR adds GetEvents and GetErrors methods . This is useful so that the watcher used in fsnotify can be more easily mocked. This does not break the existing interface it only adds two new methods.

Only two methods that should be reviewable by inspection. A compile time test is added to fsnofify_test.go that ensures the new interface is implemented.

This should not be necessary.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -13,6 +13,17 @@ import (
 	"fmt"
 )
 
+
+// GetEvents returns a channel of events.
+func (w *Watcher) GetEvents() <-chan Event {
+	return w.Events
+}
+
+// GetErrors returns a channel of errors.
+func (w *Watcher) GetErrors() <-chan error {
+	return w.Errors
+}
+
 // Event represents a single file system notification.
 type Event struct {
 	Name string // Relative path to the file or directory.
--- a/fsnotify_test.go
+++ b/fsnotify_test.go
@@ -12,6 +12,17 @@ import (
 	"time"
 )
 
+type watcher interface {
+	Close() error
+	Add(string) error
+	Remove(string) error
+	GetEvents() <-chan Event
+	GetErrors() <-chan error
+}
+
+// ensure that the Watcher meets the watcher interface.
+var _ watcher = &Watcher{}
+
 func TestEventStringWithValue(t *testing.T) {
 	for opMask, expectedString := range map[Op]string{
 		Chmod | Create: `"/usr/someFile": CREATE|CHMOD`,


iaburton:
This PR would be nice to have 🙂 

Just wanted to point out for those having a hard time mocking/testing code that uses this, a partial workaround is to use [method expression and method values](https://go.dev/ref/spec#Method_expressions) to take references to the needed funcs/methods, which you can then change in your tests:

```go
var (
	// vars for making the fsnotify.Watcher itself easier to stub out in tests
	newWatcher   = fsnotify.NewWatcher
	watcherAdd   = (*fsnotify.Watcher).Add
	watcherClose = (*fsnotify.Watcher).Close
	// etc...
)
```
Then later
```go
watcher, err := newWatcher()
// handle
err = watcherAdd(watcher, "/my/dir")
```
Note that package vars like this aren't concurrency safe, so if you have parallel tests make sure tests that change this with other tests or those that expect the real functions don't run at the same time. Or you can save these references into a higher level type and only change them out in that instance rather than for the entire package 🙂

arp242:
I can't say I'm a huge fan of the Get.. method names, and it's confusing as well since it's identical to the existing Events and Errors fields, which is not very obvious unless you look at the source.

daniel-garcia:
Would you be willing to change the method names to omit Get? That would break the existing interface but would make the object mockable.

arp242:
That would conflict with the field name, and changing it from a field to a method isn't a compatible change. It's not really the name that's the issue, it's that there are two ways to access the exact same thing.

This makes the API worse for everyone for the benefit of a very small group of people who want to mock this out for some test (which, arguably, is not necessarily a good idea in the first place).

It seems that the work-around functional? Alternatively, you can make your own little wrapper with different implementations if need be, which isn't too hard. If it would be completely impossible to mock things out, then sure, we should probably do something to address that. But I don't think it is?

The *Real Solution™* for a v2 would probably be to make the library operate on fs.FS. Actually, this is something that could perhaps be added in a compatible way in v1 as well.

daniel-garcia:
I'd like to see a proposal for a v2 api that follows the best practices you see in the standard library. Most, if not all, interactions with external processes have functions that are defined aby an interface... albeit sometimes an informal one. For example, context requires you to call a function to get a channel instead  of just providing the raw channel for that.    Its also quite easy to fork a project i suppose but i was hoping to avoid having ro maintain a fork. Putting a wrapper in another project is probably as much work as maitaining a fork for the scale of changes we are talking about.

arp242:
> Putting a wrapper in another project is probably as much work as maitaining a fork for the scale of changes we are talking about.

It's a few lines of code:

	type MyWatcher struct{ *fsnotify.Watcher }

	func (w MyWatcher) GetEvents() chan fsnotify.Event { return w.Watcher.Events }
	func (w MyWatcher) GetErrors() chan error          { return w.Watcher.Errors }
	func NewWatcher() (*MyWatcher, error) {
		if err != nil {
			return nil, err
		}
		return &MyWatcher{w}, nil
	}

daniel-garcia:
Making a wrapper is what people already do. 

https://github.com/infobloxopen/hotload/blob/main/fsnotify/fsnotify_wrapper.go


The point of the PR is to have a project dependency where thats not needed.

arp242:
Okay, well, I don't see how that is a lot of work to be honest.

If this was designed from day one then a function would obviously be better, but this package is over a decade old and not much we can do about it now. I'd really like to avoid making the API worse for a "quick fix" as a few of these accumulate and before you know it you've got a big mess like net/http.

PR #368, opened 2021-06-05
provide polling plan9 support

majiru:

Provides a very basic polling implementation for plan9. Its not fantastic but should serve the purpose for when fsnotify
is included and not used, or if its used very minimally.

qid.go

`GOOS=plan9 go build`

`go test`
on a plan9/9front machine

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-// +build !plan9
-
 // Package fsnotify provides a platform-independent interface for file system notifications.
 package fsnotify
 
--- a/qid.go
+++ b/qid.go
@@ -1,0 +1,120 @@
+// Copyright 2010 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+// +build plan9
+
+package fsnotify
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+	"time"
+)
+
+func dirStat(s string) (*syscall.Dir, error) {
+	fi, err := os.Stat(s)
+	if err != nil {
+		return nil, os.ErrNotExist
+	}
+	d, ok := fi.Sys().(*syscall.Dir)
+	if !ok || d == nil {
+		return nil, fmt.Errorf("could not cast %s to dir", s)
+	}
+	return d, nil
+}
+
+// Watcher watches a set of files, delivering events to a channel.
+type Watcher struct {
+	Events chan Event
+	Errors chan error
+
+	add    chan string
+	remove chan string
+	exit   chan struct{}
+}
+
+// NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
+func NewWatcher() (*Watcher, error) {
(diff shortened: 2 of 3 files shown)


nathany:
Thanks for the contribution.

There is talk of adding a polling fallback for other platforms. See #9.

arp242:
Right now this PR uses the Plan 9-specific `syscall.Dir`, so it won't work on other systems. Is there a specific reason for that? I plan to add polling support soon (e.g. sometime this year), and I think a generic cross-platform implementation would work on Plan 9 as well?

I spent quite a bit of time trying to get Plan 9 set up with Go and networking (9legacy), and have thus far failed to get a usable system :-( Plan 9 is quite different from what I've used before. Although I'd like to get back to this at some point, I've kind of given up on it for the time being.

Having code that *only* works on one platform with no way to test it in the CI or even locally is probably not a good idea because I have no way to test it, implement additional methods/features as they get added, etc. All I can do is check if it compiles with `GOOS=plan9`, but I don't think that's really enough.

Also, I think this implementation will only watch files and not directories, considering it calls `dirStat()` which always casts to `fi.Sys().(*syscall.Dir)` on every `Add()`?

majiru:
I apologize for the late reply, I honestly was not expecting this thread to ever get bumped. On the topic of fsnotify, I think your approach of just a fallback polling implementation sounds like the best approach. Feel free to use any of the code here for that (if you'd like), but I dont have much interest in working more on this. The reason for using the syscall.Dir, is for access to the QID version which is how you can tell if a file has been updated between polls. For plan9 specifically it doesn't make too much sense to watch directories(that would just track the stats of the dir itself, not its contents). Hence why the plan9 specific code is written this way.

On the topic of plan9, I wanted to mention that for using the system you will have a much better time trying 9front. It is much more actively developed, and has quite a bit more documentation. On the topic of running it in the CI, I have tried and given up a couple times to get 9front running within a github CI, so I have nothing to suggest there. You mention you didn't think just cross compiling with `GOOS=plan9` is enough, but I would disagree it's certainly better then nothing. A lot of the times this library is included for some specific config reloader that is not particularly needed, so as long as it builds it is generally fine.

I appreciate you taking the time to try and learn about the system, I can offer to help you get things setup in the future, if you felt the desire.

arp242:
> The reason for using the syscall.Dir, is for access to the QID version which is how you can tell if a file has been updated between polls.

Ah right, this sounds a bit like [i_version planned for Linux](https://lwn.net/Articles/905931/)?

> you will have a much better time trying 9front. It is much more actively developed, and has quite a bit more documentation

That's good to know, thanks.

> You mention you didn't think just cross compiling with GOOS=plan9 is enough, but I would disagree it's certainly better then nothing.

In my experience most bugs and problems come from runtime differences and bugs in behaviour; it all compiles fine and it runs fine, but it just sends the wrong events, events are missed, there are extraneous events, things behave different/weird in edge cases, and things like that.

At any rate, we need a generic poller backend anyway, and that should work on Plan 9 too. If there are meaningful optimisations that can be made for Plan 9 specifically – such as using the QID – then we can look into that. I'm not opposed to having Plan9-specific code, but if it can be avoided as much as possible then that's probably a win for everyone.

PR #370, opened 2021-06-25
Fix: Update watch paths when renaming directories with sub-watches on Windows 

### LINKED ISSUE #243: Creating a file inside a folder created using windows explorer generates events for the wrong dir

### LINKED ISSUE #259: Renaming folder which already added to watch

arnegroskurth:

Fixes #259
Fixes #243

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -5,6 +5,8 @@ package fsnotify
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 	"runtime"
 	"strings"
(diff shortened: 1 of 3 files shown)


windows.go
```diff
@@ -464,6 +465,17 @@ func (w *Watcher) readEvents() {
...
 				watch.rename = name
 			case syscall.FILE_ACTION_RENAMED_NEW_NAME:
+
+				// update saved path of all sub-watches
+				oldFullName := filepath.Join(watch.path, watch.rename)
+				for _, watchMap := range w.watches {
+					for _, otherWatch := range watchMap {
+						if strings.HasPrefix(otherWatch.path, oldFullName) {
+							otherWatch.path = filepath.Join(fullname, strings.TrimPrefix(otherWatch.path, oldFullName))
+						}
```
nathany, line 475: I think you need some locking when operating on the map.

nathany (commented):
Thanks for the contribution.

integration_test.go
```diff
@@ -1226,6 +1227,55 @@ func TestRemoveWithClose(t *testing.T) {
...
+			if !ok {
+				return
+			}
+			events = append(events, event)
+		}
+	}()
+
+	addWatch(t, watcher, testDir)
+
+	if err := os.Mkdir(testDir+"/dir", 0o775); err != nil {
```
mattn, line 1251: Please use filepath.Join and below.

arp242 (approved):

Issue #183, opened 2016-10-25
Maintainers wanted


nathany:
I think it makes sense to split fsnotify out into smaller platform-specific libraries that fsnotify (or alternatives) can use. Each library needs maintainers to go off and create the library with integration tests and documentation, and to handle issues and code reviews on an ongoing basis.
- [fsevents](https://github.com/fsnotify/fsevents) is one such library, currently in progress (macOS)
- inotify for Linux is another, see #173 for my thoughts on it
- Windows also needs a library
- kqueue for BSD (and macOS, iOS)
- poller for Plan9, NFS shares, Sharepoint volumes, etc.
- also FEN for Solaris

Are you interested?

related: https://github.com/pickhardt/maintainers-wanted

letter posted to forum: https://groups.google.com/forum/#!topic/golang-nuts/Ix4sg_gDLNE

nathany:
I've setup GitHub teams for each platform. If you have already contributed to fsnotify and would be like to be added to a platform-specific team, please let myself or another maintainer know (even just open an issue on the repo you would like commit access to).

zchee:
@nathany Hi,
I had plans to fork fsnotify and integrate the FSEvents because macOS's kqueue have some bugs.

BTW, I would like to confirm just in case,

> the best solution I can come up with is to split fsnotify into several platform-specific libraries

Is It means like that?
- `fsnotify/inotify`
- `fsnotify/kqueue`
- `fsnotify/fsevents`

If so and if we want to cross-platform project using fs watcher, Do we will need to import each platform packages to `_darwin.go` etc and will need to call a different API?

nathany:
@zchee Initially I'm thinking we build up repositories for https://github.com/fsnotify/inotify, etc. independent of the fsnotify package. Then start replacing the low-level bits of the fsnotify package with dependencies on those, while maintaining the same cross-platform API.

Eventually we can also substitute kqueue for fsevents on darwin.

If you'd like to help out, take a look at the https://github.com/fsnotify/fsevents project. There is an open issue with some next steps there, and I'm sure there's much more to do.

zchee:
@nathany Ah, I see. Thanks.
So, split the repository by each of the platforms, and will add an API to integrate them on the fsnotify, right?

[...]

PR #371, opened 2021-07-08
Add Solaris FEN using EventPorts from x/sys/unix

### LINKED ISSUE #12: File Events Notifications on Solaris

nshalman:

This pull request adds fsnotify support for Solaris and illumos via Solaris FEN. It is a cleanup of #263 by @isaacdavis which in turn was heavily based on #196, authored by @gfrey. I have kept both of them in the AUTHORS file and added myself as well.

The biggest change from that iteration was moving all of the cgo and unsafe code out into x/sys/unix (see [324630](https://go-review.googlesource.com/c/sys/+/324630)). Additionally, in the process of that work, I wrote a wrapper for port_getn(3c) which I have used here to allow a single call to retrieve multiple events. 

Supersedes #263
Fixes #12 

~~I squashed and rebased #263 and have preserved my development branch.
To see how this code compares to that implementation, this URL will render the diff:
https://github.com/nshalman/fsnotify/compare/rebased-solaris-fen...nshalman:illumos-fen-safer~~
This isn't really recognizable compared to the original any more. Hopefully `fen.go` having been stripped of all the cgo
is now readable without really requiring event ports / FEN expertise.

The heart of the cgo bits now [live on in x/sys/unix](https://github.com/golang/sys/blob/da31bd327af904dd4721b4eefa7c505bb3afd214/unix/syscall_solaris.go#L747-L982)

All of my testing was run on SmartOS, but the underlying event ports code includes tests that successfully ran under CI on a Solaris host. This PR also includes a GitHub action that runs an OmniOS VM to run the tests for CI.

~~Unfortunately, I've seen intermittent test failures and I haven't be able to fully identify what is happening with those.~~
I believe I've finally found all the bugs. This code is sensitive to the short sleeps in some of the integration tests, and thus depends on #422.
I've been stress testing with `echo {1..72} | xargs -n 1 | xargs -t -I {} -P 24 go test -count={}` and the VM tests which used to be flaky are now reliable (and are set to run 10 times rather than just once)

I have also done a cursory test of compiling hugo (v0.92.0) against this branch and running the tests and resulting binary.
```
hugo $ go test --count=1 github.com/gohugoio/hugo/watcher/filenotify
ok      github.com/gohugoio/hugo/watcher/filenotify     2.818s
```
Running the resulting hugo binary under truss reveals port_getn being called with arguments revealing that it's the call in fsnotify.


go.mod
```diff
@@ -3,3 +3,5 @@ module github.com/fsnotify/fsnotify
 go 1.13
 
 require golang.org/x/sys v0.0.0-20191005200804-aed5e4c7ecf9
+
+replace golang.org/x/sys => github.com/nshalman/sys v0.0.0-20210702004434-76c4cd66abf6
```
nshalman, line 7: This needs to be removed by bumping the dependency above after the Event Ports implementation lands.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.5.0 / 2021-08-28
```
nshalman, line 3: This date is made up

nshalman (commented):
I think the README also has a table indicating support and will need to be updated as well.

Is there a preference for keeping the changes all as a single commit, or should some of these changes be split into separate commits?

nathany:
Thanks for working on this.

Should #263 be closed?

nshalman:
> Should #263 be closed?

That is what I would do.

While I have your attention, any thoughts on my question:
> Is there a preference for keeping the changes all as a single commit, or should some of these changes be split into separate commits?

Thanks!

go.mod
nshalman: It landed today!

nathany:
> > Is there a preference for keeping the changes all as a single commit, or should some of these changes be split into separate commits?

Fewer commits (or squash merge) is nice, as you've done here.

.github/workflows/test.yml
```diff
@@ -59,3 +59,17 @@ jobs:
...
+    name: Test in illumos VM
+    steps:
+    - uses: actions/checkout@v2
+    - name: Test on illumos
+      id: test
+      uses: papertigers/illumos-vm@main
+      with:
+        prepare: |
+          pkg set-publisher -r -O https://pkg.omnios.org/r151036/extra extra.omnios
+          pkg install go-116
```
nathany, line 73: or 117?

nathany:
@nshalman Is there someone with FEN/Solaris experience that can provide your code review and approval?

fen.go
```diff
@@ -9,30 +9,285 @@ package fsnotify
...
+		return nil
+	}
+	close(w.done)
+	w.port.Close()
+	<-w.doneResp
 	return nil
 }
 
 // Add starts watching the named file or directory (non-recursively).
 func (w *Watcher) Add(name string) error {
```
nathany, line 91: Does this need to be renamed AddRaw to work with #289?

.github/workflows/test.yml
nshalman: As soon as it's packaged (I just checked and it isn't yet,) definitely.

fen.go
nshalman: Very likely, but I'll want to subject this to the full integration tests.
So for now, this PR should be assumed not to implement everything it would need to land after #289.

nshalman (commented):
I'll look into getting this into shape to land after #289, and I'll see if I can rustle up a reviewer with an eye to the Event Ports bits. (Maybe @tklauser??)

fen.go
```diff
@@ -9,30 +9,285 @@ package fsnotify
...
-	return nil
+	if w.isClosed() {
+		return errors.New("FEN watcher already closed")
+	}
+	if w.port.PathIsWatched(name) {
+		return nil
+	}
+	stat, err := os.Stat(name)
```
nshalman, line 98: When this function gets renamed to AddRaw as noted above, this call will need switch to `os.Lstat` to not follow symlinks.

nshalman (commented):
I did the testing and found the need to switch the call in AddRaw from `os.Stat` to `os.Lstat` but there are two other places where I call out to `os.Stat` and I'm not 100% clear on the semantics in those situations (in `Remove` and at the end of `handleEvent`)

I'm not totally convinced that the integration tests are exercising those code paths both with and without symlinks and might need to write some additional tests to figure out for sure if they need to be changed as well or not.

My gut says that they need to be changed too, and that the passing tests when only changing the call in AddRaw indicates a gap in the testing.

fen.go
nathany: Good catch. Are there uses of Stat on other OSes that should be changed to Lstat in #289?

fen.go
nathany: FYI: #289 is merged

fen.go
nshalman: I haven't looked. I *think* the use of Stat here is somewhat specific to how file watching is done with Event Ports (where you pass in the stat struct rather than just a path.)

My plan is to add some tests that confirm that Lstat is needed in the other two locations as well (under the assumption that `Add` is fully in charge of the dereferencing of a symlink) and see if they pass or fail on the other operating systems.

It occurs to me that I suddenly have questions about how Remove is supposed to work if a path was watched with Add vs with AddRaw. Probably need some more tests around that too...

.github/workflows/test.yml
nshalman: Updated.

fen.go
nshalman: Done, but with the concerns raised in #387 / #388

fen.go
nshalman: I did look. No other implementation uses os.Stat. Concerns moved to the freshly opened issue and PR.

nathany:
@nshalman It may be better to build this out as a separate FEN project -- e.g. `fsnotify/fen`

I can't see myself maintaining or even doing code reviews on this -- I have enough trouble keeping up with the existing platforms. 

What do you think?

nathany:
I think there is the potential to eventually make fsnotify utilize an external FEN library. 

In the interim, people who really want Solaris support could wrap fsnotify + fen libraries in some way. For example, Hugo uses both fsnotify + polling (I admit I haven't looked too closely at how they are doing it).
https://github.com/gohugoio/hugo/tree/master/watcher/filenotify

The idea of making fsnotify a very light wrapper around platform-specific libraries is something I've been thinking about for a long time. A large part of that is to have small teams that own each platform-specific library and can work freely on them.

The cross-platform aspect makes it really tricky for contributors, so I'd like to find a way to alleviate those troubles if possible.

jclulow:
> 


> The cross-platform aspect makes it really tricky for contributors, so I'd like to find a way to alleviate those troubles if possible.

At it's heart, `fsnotify` is a cross-platform abstraction layer, so it feels like that's what contributors are signing up for when they contribute.  Indeed, people have been trying to add support for illumos to this library for years now:

| When | PR | Link |
| ------ | ----- | ---- |
| 5th February 2017 | Add support for solaris FEN |  #196 |
| 28th August 2018 | add Solaris FEN support | #263 |
| 7th July 2021 | Add Solaris FEN using EventPorts from x/sys/unix |  #371 |

> The idea of making fsnotify a very light wrapper around platform-specific libraries is something I've been thinking about for a long time. A large part of that is to have small teams that own each platform-specific library and can work freely on them.

That sounds great, but it's not the status quo today.  I think it would be unfortunate to put something that might never happen on the critical path to getting this change integrated so people can make use of it.

> In the interim, people who really want Solaris support could wrap fsnotify + fen libraries in some way. For example, Hugo uses both fsnotify + polling (I admit I haven't looked too closely at how they are doing it). https://github.com/gohugoio/hugo/tree/master/watcher/filenotify

This would create a lot of extra technical and social work: rather than amend the cross-platform abstraction library, you are asking just one platform to go and reimplement that abstraction in any number of other bodies of consuming software.  It seems like `fsnotify` is popular; that could be tens or hundreds of other bodies of software, each requiring rework and negotiation with maintainers.

I agree that we, and you, are all already resource constrained enough, which is why people have been trying to do the work just once, in the right place: here.

It isn't clear how much or what sort of additional code review is necessary before a merge.  Can you outline your concerns, and what you feel remains outstanding?  People in the community can likely assist with maintenance and debugging, say¸ as issues arise in future -- but it really needs to get merged so that people are actually using it in downstream software.  Right now it's just sitting in a branch.

nathany:
Joshua

For the past year I have been seriously considering archiving this project and calling the whole thing quits. I have literally been looking for maintainers to help with it for **over 5 years now** (#183). 

The last time I merged someone's code it had several regressions. I'm not suggesting that will be a problem with this PR, but it is very difficult to maintain this project and nobody else wishes to step up to do it.

If @nshalman wants to fix the conflicts and merge this in, that's up to him. I will not be the one to hit merge, nor the one to take responsibility for maintaining it.

nathany:
Thanks for the push. #413

nshalman:
I said forever ago:
> I'll look into getting this into shape to land after #289, and I'll see if I can rustle up a reviewer with an eye to the Event Ports bits. (Maybe @tklauser??)

Given that #289 was backed out and that I'm paying attention again, maybe we can get this reviewed (not by you @nathany, you've done so much already and I'm very grateful.)
I'm curious if @tklauser can review for anything obviously wrong.
And perhaps @toasterson can take a look or perhaps even take the code for a spin?

I've moved the illumos VM test into what I hope is a "slow tests" that can be triggered manually and to which we could hopefully also add a similar FreeBSD test.

> If @nshalman wants to fix the conflicts and merge this in, that's up to him. I will not be the one to hit merge, nor the one to take responsibility for maintaining it.

I would like to merge this before the project is archived if that ends up being the route chosen, so I will try to get it in shape fast enough for that possibility, but as I have mentioned elsewhere, I do intend to stick around and maintain this particular port, and help with the project in general if it remains active.

fen.go
nshalman: Unresolving this to draw the attention of other reviewers. I *think* this is fine and correct, but it's worth close attention.

Toasterson:
@nshalman I can give it a spin/review but it will take me some days to get some free cycles. Mattermost and hugo are ones I can think of we could distribute patched versions to run. Especially as omnios does this already with the mattermost they package.

nshalman:
> Especially as omnios does this already with the mattermost they package.

Wow, so the code from #263 was already in use: 
https://github.com/omniosorg/omnios-extra/blob/687e07dc8b339613cb888b29aab36c43b91d4ed2/build/mattermost/patches-mattermost-server/fen.go.patch

Good to know.

tklauser:
@nshalman I'll try to review as well once I have some free cycles.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## v1.6.0 / 2022-01-28
```
nshalman, line 10: Once again, this date is made up and will need to be settled before merging...

LICENSE
```diff
@@ -1,5 +1,5 @@
 Copyright (c) 2012 The Go Authors. All rights reserved.
-Copyright (c) 2012-2019 fsnotify Authors. All rights reserved.
+Copyright (c) 2012-2021 fsnotify Authors. All rights reserved.
```
nshalman, line 2: 2022...

fen.go
```diff
@@ -9,30 +9,285 @@ package fsnotify
...
 
 // Watcher watches a set of files, delivering events to a channel.
 type Watcher struct {
 	Events chan Event
 	Errors chan error
+
+	port *unix.EventPort
+
+	done     chan struct{} // Channel for sending a "quit message" to the reader goroutine
+	doneResp chan struct{} // Channel to respond to Close
```
nshalman, line 27: I'm not convinced this channel actually gets used properly??

go.mod
```diff
@@ -2,6 +2,6 @@ module github.com/fsnotify/fsnotify
 
 go 1.16
 
-require golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
+require golang.org/x/sys v0.0.0-20210817134402-fefb4affbef3
```
nshalman, line 5: This is a minimum version to pick up the event ports code. Should we update to the latest? Some other specific version e.g. one used in a particular release of golang?

.github/workflows/slow-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: macos-latest
+    name: Test in illumos VM
+    steps:
+    - uses: actions/checkout@v2
+    - name: Test on illumos
+      id: test
+      uses: papertigers/illumos-vm@main
+      with:
+        prepare: |
+          pkg set-publisher -r -O https://pkg.omnios.org/r151036/extra extra.omnios
```
nshalman, line 14: @papertigers Do you include the extra repo in the latest images so that I could skip this part?
Alternately is there a clever way to make this more generic to survive across releases?

.github/workflows/slow-tests.yml
```diff
@@ -0,0 +1,17 @@
+name: slow-tests
+on: [workflow_dispatch]
```
nshalman, line 2: Example of a run manually triggered in my fork: https://github.com/nshalman/fsnotify/actions/runs/1708324643

fen.go
nshalman: Removing it doesn't break anything...

nshalman:
Just leaving a breadcrumb for myself based on an IRC conversation. The errors I'm seeing in CI suggest that when the VM is slowing down the code, events are getting batched together. `handleEvent` in its current state assumes that only one event should match the events bitmap, but clearly that doesn't have to be the case. I think handleEvent needs to be modified to be able to emit multiple fsnotify events for a single event ports event.

Also, I think it would be possible to trigger this state more reliably by altering the call to `port_getn` to take a timeout and a minimum greater than 1.

More work remains, but at least I think I understand what it is!

nshalman:
In order to stress test before and after, I wrote a tweak that I think increases the chance of batching and does seem to make the original test failures occur more frequently. For posterity, it was:
```diff
diff --git a/fen.go b/fen.go
index 80f75e7..5d8c7c0 100644
--- a/fen.go
+++ b/fen.go
@@ -130,9 +130,11 @@ func (w *Watcher) readEvents() {
        defer close(w.Events)

        pevents := make([]unix.PortEvent, 8, 8)
+       timeout := new(unix.Timespec)
+       timeout.Nsec = 90000
        for {
-               count, err := w.port.Get(pevents, 1, nil)
-               if err != nil {
+               count, err := w.port.Get(pevents, 8, timeout)
+               if err != nil && err != unix.ETIME{
                        // Interrupted system call (count should be 0) ignore and continue
                        if err == unix.EINTR && count == 0 {
                                continue
```

CHANGELOG.md
nathany: yes, you may want to do the release as a separate commit/PR, and just add to the unreleased section in this PR

there may be other things in the 1.6.0 release

go.mod
nathany: probably latest? if everything still passes in our test matrix

nshalman:
With #422 incorporated, along with the changes in 04e3205 I'm no longer able to trigger test failures.
I now trust this code and believe it to be superior to #263 from a correctness perspective.

I imagine there is room for improvement that my reviewers with hopefully provide when they find time, but this no longer needs to be marked as draft. I would feel comfortable supporting this code into the future if it was merged today.

nshalman:
Again, with extensive stress testing I was able to trigger a test failure which led me to realize this issue in my x/sys/unix Event Ports implementation:
https://go-review.googlesource.com/c/sys/+/380034

So, I need to land that, integrate it here, and stress test it all again.

.github/workflows/slow-tests.yml
nshalman: Following up, @papertigers was kind enough to do this for the @r38 branch using a newer OmniOS release:
https://github.com/papertigers/fsnotify/commit/2db24b5fb78a888b94c923fa9e6e9b08d8abc554

I will pull in those changes soon.

tnorlin:
I am running Kubernetes on illumos and your changes looks good so far for me! 
I used to have this errors:

> E0130 08:57:42.227066    4762 dynamic_cafile_content.go:165] "Failed to watch CA file, will retry later" err="error creating fsnotify watcher: FEN based watcher not yet supported for fsnotify\n"

But they are gone now.

nshalman:
Now that https://go-review.googlesource.com/c/sys/+/380034 has been merged, I hope to pick this up again some time soon.

arp242:
I created a branch with main merged in over here, FYI: https://github.com/fsnotify/fsnotify/commits/fen-v2-upd

Mostly just boring changes to the README and such; I tried to make a PR for it to merge it in to here, but GH won't let me :-/

I put the illumos test runner in main already.

Some comments:

- In Remove() it calls `os.Stat()` and then calls handleDirectory() or DissociatePath(). I think it might be better to keep track of what we've associated in the watcher, rather than relying on os.Stat() and the filesystem state.

  For example, what happens if we're watching `path` and someone does:

	rm -rf path && touch path

  This now changed from a directory to a file (or the reverse), and if this happens at roughly the same time as the Watcher.Remove() call (not inconceivable) the stat call may return something different from what we expect.

  This has been (and still is) a source of issues with other backends anyway. In general, I think accessing file information "directly" through ReadDir(), Stat(), etc. should be minimized and that we should prefer to keeping an internal state of file descriptors and the like.

- In Remove() it calls handleDirectory() with the dissociateFile callback, which reads the directory and calls it for every file. I think this would be racy in the same way: someone calls Watcher.Remove() and at the same time files are removed, so they won't get picked up by the readdir, and it will never be disassociated.

- In handleEvent() it only sends a CHMOD if the file mode bits changes, but I think we also want this for user/group change? It's not clear to me what events are being filtered here.

- Do we really need to do a os.Stat() and re-associate the file in handleEvent()?

  From the [port_associate] docs, "When the event for the associated file_obj is retrieved, the object is no longer associated with the port", but it's probably better to keep a state of things internally rather than calling stat all the time?

  I assume that FEN already keeps the file descriptors open, so we might as well hang on to them ourselves too. stat() is a comparatively expensive call.

  [port_associate]: https://illumos.org/man/3C/port_associate

arp242:
Oh, and the tests fail; output with some comments added.

---

The sub directory is not watched; we should not see changes to /sub/file2.

	--- FAIL: TestWatch (0.00s)
		--- FAIL: TestWatch/subdir (1.01s)
			helpers_test.go:33: 
				have:
					CREATE               "/sub"
					CREATE               "/file"
					CREATE               "/sub/file2"
					REMOVE               "/sub"
					REMOVE               "/sub/file2"
					REMOVE               "/file"
				want:
					CREATE               "/sub"
					CREATE               "/file"
					REMOVE               "/sub"
					REMOVE               "/file"


---

We can add a illumos block for these; renaming or removing a watched directory behaves different for every backend. Maybe we should work on unifying this at some point, but I don't know if that's even possible.

The cyclic symlink one is also one that behaves pretty different on every backend, with macOS being completely borked 🤷

	--- FAIL: TestWatchRename (0.00s)
		--- FAIL: TestWatchRename/rename_watched_directory (0.83s)
			helpers_test.go:33: 
				have:
					CREATE               "/dir"
					RENAME               "/dir"
					CREATE               "/dir-renamed"
					CREATE               "/dir-renamed/file"
				want:
					CREATE               "/dir"
					RENAME               "/dir"
					CREATE               "/dir-renamed"
					RENAME               "/dir"
					CREATE               "/dir/file"
	--- FAIL: TestWatchSymlink (0.00s)
		--- FAIL: TestWatchSymlink/cyclic_symlink (0.74s)
			helpers_test.go:33: 
				have:
					REMOVE               "/link"
					CREATE               "/link"
					WRITE                "/link"
				want:
					WRITE                "/link"
					CREATE               "/link"
	--- FAIL: TestWatchRm (0.00s)
		--- FAIL: TestWatchRm/remove_watched_directory (0.62s)
			helpers_test.go:33: 
				have:
					REMOVE               "/"
					REMOVE               "/file"
				want:
					REMOVE               "/file"
					REMOVE|WRITE         "/"

nshalman:
@arp242 Thank you for taking a look at this. My biggest concern at the moment is that under stress tests I've been able to trigger the panic in my Event Ports code in x/sys/unix which shouldn't be possible so I'm worried that there's still a bug in there somewhere.

I can certainly take a pass at rebasing this onto the latest code in the main branch, and as for your comments (thank you for those!!) I will try to find some time to look into them.

arp242:
> I can certainly take a pass at rebasing this onto the latest code in the main branch

Merging is a little bit easier, and with "squash and merge" it shows up as one commit in the main branch and we can rewrite the commit message. I mean, you can rebase if you want, but it's not really needed: can just `git merge` my branch and push it. I can push it here too if you want?

I set up a bunch of QEMU machines for testing over the weekend; I haven't done illumos yet but I'll get it set up and see if I can reproduce and maybe fix that issue, as well as the failing `TestWatch/subdir` test.

arp242:
Also, should maybe look at setting up a Solaris runner too; I don't think Solaris and illumos have shared much (or any?) code since OpenSolaris got Oracle'd?

Can maybe use https://github.com/vmactions/solaris-vm for this; I looked at that a few weeks ago, but it didn't seem to have an easy way to install Go, so need to look at that.

nshalman:
> Merging is a little bit easier, and with "squash and merge" it shows up as one commit in the main branch and we can rewrite the commit message. I mean, you can rebase if you want, but it's not really needed: can just `git merge` my branch and push it. I can push it here too if you want?

Done. I can confirm that the tests are broken for me.

nshalman:
> Oh, and the tests fail; output with some comments added.
> 
> The sub directory is not watched; we should not see changes to /sub/file2.
> 
> ```
> --- FAIL: TestWatch (0.00s)
> 	--- FAIL: TestWatch/subdir (1.01s)
> 		helpers_test.go:33: 
> 			have:
> 				CREATE               "/sub"
> 				CREATE               "/file"
> 				CREATE               "/sub/file2"
> 				REMOVE               "/sub"
> 				REMOVE               "/sub/file2"
> 				REMOVE               "/file"
> 			want:
> 				CREATE               "/sub"
> 				CREATE               "/file"
> 				REMOVE               "/sub"
> 				REMOVE               "/file"
> ```

I don't currently know how to not watch the subdirectory. It might require surgery in the x/sys/unix implementation.
If I'm watching a directory and a file is created inside, I watch it. If instead of a file it happens to be a directory and I watch that directory, I will get notified if something gets created inside the sub directory.

nshalman:
> Also, should maybe look at setting up a Solaris runner too; I don't think Solaris and illumos have shared much (or any?) code since OpenSolaris got Oracle'd?
> 
> Can maybe use https://github.com/vmactions/solaris-vm for this; I looked at that a few weeks ago, but it didn't seem to have an easy way to install Go, so need to look at that.

My understanding is that the Event Ports code hasn't diverged much if at all between Solaris and illumos since the fork. The x/sys/unix bits I added are in the solaris files and are available on both.

arp242:
> My understanding is that the Event Ports code hasn't diverged much if at all between Solaris and illumos since the fork. The x/sys/unix bits I added are in the solaris files and are available on both.

My concern is that it might behave subtly different; for example the macOS and BSD implementations all use kqueue, and if you read the man pages on each system you'd think it was identical because they're near copies and API-compatible, but in practice they're not entirely identical and can behave subtly different.

For example [this PR](https://github.com/fsnotify/fsnotify/pull/262#issuecomment-1201655890) works great on macOS and OpenBSD, but completely breaks FreeBSD and NetBSD; if we had merged it, it would have been a major regression. There are some other test cases where one or more kqueue platforms behave different.

Alternatively, we can also just say "we support illumos, and Solaris might work but it's not tested" if it proves to be too hard to add a Solaris test runner; I'd be fine with that too. But I do think it's important we have test runners for platforms we claim to support.

arp242:
I can't reproduce these consistently, but I encountered the following panics with "go test":

	goroutine 96 [running]:
	github.com/fsnotify/fsnotify.(*Watcher).Close(0xc000013e60)
			/home/martin/fsnotify/fen.go:87 +0x56
	created by github.com/fsnotify/fsnotify.TestClose.func4
			/home/martin/fsnotify/integration_test.go:472 +0x185

And:

	panic: mismanaged memory

	goroutine 679 [running]:
	golang.org/x/sys/unix.(*EventPort).peIntToExt(0xc0002311d0, 0xc000234600, 0xc00014cda8)
			/home/martin/go/pkg/mod/golang.org/x/sys@v0.0.0-20220412211240-33da011f77ad/unix/syscall_solaris.go:958 +0x258
	golang.org/x/sys/unix.(*EventPort).Get(0xc0002311d0, {0xc00014cda8, 0x8, 0xc000032548?}, 0x40bffe?, 0xfffffc7fef16cc20?)
			/home/martin/go/pkg/mod/golang.org/x/sys@v0.0.0-20220412211240-33da011f77ad/unix/syscall_solaris.go:1001 +0x2e9
	github.com/fsnotify/fsnotify.(*Watcher).readEvents(0xc0001a1e60)
			/home/martin/fsnotify/fen.go:142 +0xe5
	created by github.com/fsnotify/fsnotify.NewWatcher
			/home/martin/fsnotify/fen.go:47 +0x12a

I'm running OmniOS r151042 in a QEMU VM, with Go 1.18:

	% uname -a
	SunOS omnios 5.11 omnios-r151042-2a01bcc289 i86pc i386 i86pc

	% go version
	go version go1.18.4 illumos/amd64

The first likely isn't related to some QEMU-specific thing, but the second one *could* be, although it's *probably* a bug in golang.org/x/sys/unix/syscall_solaris.go

arp242:
This patch fixes the subdir test; the problem was that in `handlEvent()` it will use the `w.port.PathIsWatched()` syscall to check if a path is watched, and watch all files in directories if it is. But this isn't correct: we have two "watchers": the one the user added via `Watcher.Add()` (we should list those and setup watchers), and those that we created (we *shouldn't* list the directory for those!)

So keep our own `Watcher.watches` map for this.

This broke the `TestWatchRename/rename_watched_directory` test though, but I think that's okay: the behaviour of that is already odd and inconsistent on different platforms. As far as I'm concerned we don't need to fix that in this PR.

With this, all tests pass (but still those panics remaining, and previous comments about race conditions):

```
[~cg/fsnotify](fen-v2)% goon illumos go test
goon: illumos already started; nothing to do
PASS
ok      github.com/fsnotify/fsnotify    7.416s
````

The patch is a bit ugly and not very cleaned up, but should hopefully illustrate the problem and direction towards a solution:

```
diff --git fen.go fen.go
index 792c419..99adf46 100644
--- fen.go
+++ fen.go
@@ -13,13 +13,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"golang.org/x/sys/unix"
 )
 
-const eventBits = unix.FILE_MODIFIED | unix.FILE_ATTRIB | unix.FILE_NOFOLLOW
-
 // Watcher watches a set of files, delivering events to a channel.
 type Watcher struct {
 	Events chan Event
@@ -27,8 +26,9 @@ type Watcher struct {
 
 	done chan struct{} // Channel for sending a "quit message" to the reader goroutine
 
-	mu   sync.Mutex
-	port *unix.EventPort
+	mu    sync.Mutex
+	paths map[string]struct{}
+	port  *unix.EventPort
 }
 
 // NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
@@ -38,6 +38,7 @@ func NewWatcher() (*Watcher, error) {
 	w := new(Watcher)
 	w.Events = make(chan Event)
 	w.Errors = make(chan error)
+	w.paths = make(map[string]struct{})
 	w.port, err = unix.NewEventPort()
 	if err != nil {
 		return nil, err
@@ -97,6 +98,15 @@ func (w *Watcher) Add(name string) error {
 	if w.isClosed() {
 		return errors.New("FEN watcher already closed")
 	}
+
+	w.mu.Lock()
+	if _, ok := w.paths[name]; ok {
+		w.mu.Unlock()
+		return nil
+	}
+	w.paths[name] = struct{}{}
+	w.mu.Unlock()
+
 	if w.port.PathIsWatched(name) {
 		return nil
 	}
@@ -195,6 +205,32 @@ func (w *Watcher) handleDirectory(path string, stat os.FileInfo, handler func(st
 	return handler(path, stat)
 }
 
+func dbgEvent(name string, mask int32) {
+	names := []struct {
+		n string
+		m int32
+	}{
+		{"FILE_ACCESS", unix.FILE_ACCESS},
+		{"FILE_MODIFIED", unix.FILE_MODIFIED},
+		{"FILE_ATTRIB", unix.FILE_ATTRIB},
+		{"FILE_TRUNC", unix.FILE_TRUNC},
+		{"FILE_NOFOLLOW", unix.FILE_NOFOLLOW},
+		{"FILE_DELETE", unix.FILE_DELETE},
+		{"FILE_RENAME_TO", unix.FILE_RENAME_TO},
+		{"FILE_RENAME_FROM", unix.FILE_RENAME_FROM},
+	}
+	var l []string
+	for _, n := range names {
+		if mask&n.m == n.m {
+			l = append(l, n.n)
+		}
+	}
+	if i := strings.Index(name, "/Test"); i > -1 {
+		name = name[i:]
+	}
+	fmt.Printf("%-50s → %s (0x%x)\n", name, strings.Join(l, " | "), mask)
+}
+
 // handleEvent might need to emit more than one fsnotify event
 // if the events bitmap matches more than one event type
 // (e.g. the file was both modified and had the
@@ -205,6 +241,8 @@ func (w *Watcher) handleEvent(event *unix.PortEvent) error {
 	path := event.Path
 	fmode := event.Cookie.(os.FileMode)
 
+	//dbgEvent(path, events)
+
 	var toSend *Event
 	reRegister := true
 
@@ -256,8 +294,13 @@ func (w *Watcher) handleEvent(event *unix.PortEvent) error {
 
 	if events&unix.FILE_MODIFIED == unix.FILE_MODIFIED {
 		if fmode.IsDir() {
-			if err := w.updateDirectory(path); err != nil {
-				return err
+			w.mu.Lock()
+			_, ok := w.paths[path]
+			w.mu.Unlock()
+			if ok {
+				if err := w.updateDirectory(path); err != nil {
+					return err
+				}
 			}
 		} else {
 			toSend = &Event{path, Write}
@@ -282,6 +325,8 @@ func (w *Watcher) handleEvent(event *unix.PortEvent) error {
 }
 
 func (w *Watcher) updateDirectory(path string) error {
+	//fmt.Println("\tupdateDirectory:", path)
+
 	// The directory was modified, so we must find unwatched entities and
 	// watch them. If something was removed from the directory, nothing will
 	// happen, as everything else should still be watched.
@@ -313,6 +358,7 @@ func (w *Watcher) associateFile(path string, stat os.FileInfo) error {
 	if w.isClosed() {
 		return errors.New("FEN watcher already closed")
 	}
+	//fmt.Println("\t\tassociateFile:", path)
 	// This is primarily protecting the call to AssociatePath
 	// but it is important and intentional that the call to
 	// PathIsWatched is also protected by this mutex.
@@ -333,8 +379,10 @@ func (w *Watcher) associateFile(path string, stat os.FileInfo) error {
 			return err
 		}
 	}
-	fmode := stat.Mode()
-	return w.port.AssociatePath(path, stat, eventBits, fmode)
+
+	return w.port.AssociatePath(path, stat,
+		unix.FILE_MODIFIED|unix.FILE_ATTRIB|unix.FILE_NOFOLLOW,
+		stat.Mode())
 }
 
 func (w *Watcher) dissociateFile(path string, stat os.FileInfo) error {
diff --git integration_test.go integration_test.go
index 4533e36..c7787ef 100644
--- integration_test.go
+++ integration_test.go
@@ -195,11 +195,6 @@ func TestWatchRename(t *testing.T) {
 				RENAME       "/dir"                 # mv
 				CREATE       "/dir-renamed"
 				CREATE       "/dir-renamed/file"    # touch
-			solaris:
-				CREATE       "/dir"                 # mkdir
-				RENAME       "/dir"                 # mv
-				CREATE       "/dir-renamed"
-				CREATE       "/dir-renamed/file"    # touch
 
 			# TODO: no results for the touch; this is probably a bug; windows
 			# was fixed in #370.
@@ -207,6 +202,12 @@ func TestWatchRename(t *testing.T) {
 				CREATE               "/dir"           # mkdir
 				CREATE               "/dir-renamed"   # mv
 				REMOVE|RENAME        "/dir"
+
+			# TODO: same here
+			solaris:
+				CREATE       "/dir"                 # mkdir
+				RENAME       "/dir"                 # mv
+				CREATE       "/dir-renamed"
 		`},
 	}
 
```

**Comment by nshalman** [comment]
I've added a change based on your suggestions. Tests now pass except for when I get failures like this:

``` 
~/fsnotify$ go test

goroutine 319 [running]:
github.com/fsnotify/fsnotify.(*Watcher).Close(0xc000384210)
        /home/nshalman/fsnotify/fen.go:87 +0x5b
created by github.com/fsnotify/fsnotify.TestClose.func4
        /home/nshalman/fsnotify/integration_test.go:464 +0x139
exit status 2
FAIL    github.com/fsnotify/fsnotify    3.877s
```

And to be very very clear, the Event Ports code in x/sys/unix was written specifically to support fsnotify, so if it has bugs, they will probably only be found by exercising them with fsnotify. I don't think much of anything else is using it yet (I'm using it very very lightly in wireguard-go, but probably not in a way that would find the bugs)

**Comment by nshalman** [comment]
I think I've fixed the race condition on Close and I've merged main again.

**Comment by nshalman** [comment]
> And to be very very clear, the Event Ports code in x/sys/unix was written specifically to support fsnotify, so if it has bugs, they will probably only be found by exercising them with fsnotify. I don't think much of anything else is using it yet (I'm using it very very lightly in wireguard-go, but probably not in a way that would find the bugs)

I can still trigger the panic sometimes with, e.g.
```
go test --failfast -count=20 -coverprofile=cover.txt
```

**Comment by arp242** [comment]
With the cmd/fsnotify CLI tool you can reproduce the issue fairly easily:

	% mkdir /tmp/xxx
	% go run ./cmd/fsnotify /tmp/xxx

And then in another terminal:

	% for i in {1..500}; touch /tmp/xxx/$i

Seems to have a ~100% "success" rate with a loop of 500 items; can increase it a bit if that's not enough.

In my testing I found that it will *only* panic when you're sending a bunch of CREATE or REMOVE events; if you already have those 1000 files and touch them or write to them then it'll be fine, so be sure to clean the files afterwards:

	% for i in {1..500}; touch /tmp/xxx/$i; rm -f /tmp/xxx/*

I looked a bit at the sys code, but not immediatly clear to me where it's going
wrong.

**Comment by nshalman** [comment]
I've opened https://github.com/golang/go/issues/54254 to track (something I probably should have done ages ago...)

**Comment by nshalman** [comment]
Ugh. #374 will probably mean I need to make some changes to x/sys/unix to get this working again.

**Comment by nshalman** [comment]
There are still some bugs, I've found most of them. In addition to uncovering https://www.illumos.org/issues/14898 I've also found another edge case in the x/sys/unix Event Ports code thanks to the `TestClose` tests.

I need to get some fixes upstream into x/sys/unix before this can move forward again.
Thanks for both your help getting this moving, and your patience while I tackle all these other bugs elsewhere. :)

**Comment by arp242** [comment]
Yeah sure, no hurries.

Aside: the other day I found out illumos actually supports inotify too(!) Although that's not exposed in x/sys/unix either.

**Comment by jclulow** [comment]
> Aside: the other day I found out illumos actually supports inotify too(!) Although that's not exposed in x/sys/unix either.

FYI: inotify support is not universal, it's only in some of the forks.  Where it exists, it's also a compatibility shim rather than the native interface we generally prefer.

**Comment by arp242** [comment]
Ah right; was just surprised to see it at all.

**Comment by Toasterson** [comment]
We do have compat shims around for inotify, polkit as they are popular API's and thus a good place to lie our asses of when we want many applications to support us without them doing porting. But we try to eventually get porting efforts going. This is one of them :)

**Comment by arp242** [comment]
It seems the CI stopped running for this PR? 🤔 Last run was e6d3c16 8 days ago; probably relates to my changes in #484, but I haven't had any issues in other branches/PRs, so not sure why it's not working here?

**Comment by nshalman** [comment]
> It seems the CI stopped running for this PR? 🤔 Last run was [e6d3c16](https://github.com/fsnotify/fsnotify/commit/e6d3c16047b1ffe44b498d8bc614412c5775458f) 8 days ago; probably relates to my changes in #484, but I haven't had any issues in other branches/PRs, so not sure why it's not working here?

Might be because I marked it as draft.

**Comment by nshalman** [comment]
I also consider this blocked on https://go-review.googlesource.com/c/sys/+/422338 which adds `WatchedPaths()` to the underlying EventPort which we can use for the new `WatchList()` functionality. I have changes not yet pushed up that incorporate that.

**Comment by arp242** [comment]
> Might be because I marked it as draft. EDIT: Oh, it's not currently marked as draft after all... ?!

Yeah, I marked it as "ready" to see if it would fix anything, but it didn't. Plus my recursive inotify PR is draft and that seems to work.

**Comment by arp242** [comment]
> I also consider this blocked on https://go-review.googlesource.com/c/sys/+/422338 which adds WatchedPaths() to the underlying EventPort which we can use for the new WatchList() functionality. I have changes not yet pushed up that incorporate that.

The current behaviour of `WatchList()` seems fine? It should return the paths you explicitly added with Add(), which is what it does now.

The `EventPort.WatchedPaths` you added in that commit seems to return all paths associated with a port, but that's not quite right.

**Comment by nshalman** [comment]
> The current behaviour of `WatchList()` seems fine? It should return the paths you explicitly added with Add(), which is what it does now.
> 
> The `EventPort.WatchedPaths` you added in that commit seems to return all paths associated with a port, but that's not quite right.

This explanation is really helpful! And yes, based on that difference being explicit then we don't need to query the EventPort because it would naturally have a different view of things. I can even remove WatchedPaths from that EventPort CR. Thank you!

**Comment by arp242** [comment]
> I can even remove WatchedPaths from that EventPort CR. Thank you!

If it already works right now, I'd keep it. At the very least it'll be useful for testing: "are we keeping any associations around that we shouldn't? Are they properly removed up in various conditions (Watcher.Remove(), errors, etc.)?"

WatchList() was added earlier this year and it was pretty broken/inconsistent on different platforms: on Windows and Linux it would behave as described above, and on kqueue it would return everything it had a file descriptor for. To be honest I never would have merged that PR as it was, but I think Nathan was already really tired/exasperated from maintaining this "so sure, whatever I guess" (no judgement, I can relate). Anyhow, I changed the definition to what I mentioned above as that seemed the most useful one to me. Should probably tweak the comment a bit, too, as it's ambiguous now.

Another reason to keep it is that we might add a `List(all bool)` in the future (e.g. v2) where the flag controls if it should return everything we have an association or file descriptor for, or just what was explicitly added.

**Comment by arp242** [comment]
The `TestWatch/watch_a_symlink_to_a_file` I added in #498 (and merged here too) seems to generate a lot of Write events:

    --- FAIL: TestWatch/watch_a_symlink_to_a_file (0.76s)
        fsnotify_test.go:214: 
            have:
            	WRITE                "/link"
            	WRITE                "/link"
            	[.. 16578 more elided ..]
            	WRITE                "/link"
            	WRITE                "/link"
            want:
            	WRITE                "/link"

**Comment by nshalman** [comment]
With the closing of https://github.com/golang/go/issues/54363 we should be clear to pull in the latest commit of x/sys/unix and get this work unblocked again!

**Comment by arp242** [comment]
Updated it in the main branch; didn't merge it here.

So if I understand things correctly people will need a fairly recent version of illumos or it will panic? Not sure which versions of illumos distributions contain this change yet (if any?) Need to update the CI and my local test environment, as well as communicate this clearly to users.

**Comment by Toasterson** [comment]
Right now I can find the git hash `abb88ab1b9516b1ca12094db7f2cfb5d91e0a135` in omnios bloody, OpenIndiana hipster (upstream illumos-gate), and the opensolaris porting project.  I can't find it in illumos-joyent but there it might have a different hash as last gate merge was 5 hours ago, so unless it was specifically skipped it's in all upstrem tracking distros. In whcih release it made it already would be a question for @danmcd and @citrus-it as I do not know how to track it beyond scrubbing for it in github.

**Comment by danmcd** [comment]
It's in illumos-omnios:master, which will eventually land in a bloody build:

```
nowhere(~/ws/illumos-omnios)[0]% git branch
* master
  nfs-zone
  r151030
  r151032
  r151034
  r151036
  r151038
  r151042
nowhere(~/ws/illumos-omnios)[0]% git log -1 abb88ab1b9516b1ca12094db7f2cfb5d91e0a135
commit abb88ab1b9516b1ca12094db7f2cfb5d91e0a135
Author: Robert Mustacchi <rm@fingolfin.org>
Date:   Sun Aug 7 20:20:53 2022 +0000

    14898 port_associate PORT_SOURCE_FILE doesn't update user obj properly
    Reviewed by: Nahum Shalman <nahamu+illumos@gmail.com>
    Reviewed by: Andy Fiddaman <illumos@fiddaman.net>
    Approved by: Richard Lowe <richlowe@richlowe.net>
nowhere(~/ws/illumos-omnios)[0]% git branch -a --contains abb88ab1b9516b1ca12094db7f2cfb5d91e0a135
* master
  remotes/origin/HEAD -> origin/master
  remotes/origin/master
nowhere(~/ws/illumos-omnios)[0]% 
```

**Comment by danmcd** [comment]
And FWIW, you need a fairly recent SmartOS build for this fix:
```
kebe(~/ws/ij-cr)[0]% git branch -a --contains abb88ab1b9516b1ca12094db7f2cfb5d91e0a135
* master
  remotes/origin/HEAD -> origin/master
  remotes/origin/OS-8403
  remotes/origin/master
  remotes/origin/release-20220825
  remotes/origin/release-20220908
kebe(~/ws/ij-cr)[0]% 
```

**Comment by nshalman** [comment]
> So if I understand things correctly people will need a fairly recent version of illumos or it will panic? Not sure which versions of illumos distributions contain this change yet (if any?) Need to update the CI and my local test environment, as well as communicate this clearly to users.

It *might* panic the golang application (not the operating system.) For most uses cases it probably won't. It took a fair amount of stress testing for me to trip the bug.... If it does panic, the error message points explicitly at https://github.com/golang/go/issues/54254 which will guide users to look into an OS update.

@papertigers did ping me on IRC about the need for an update for the CI OS image. Might take a little bit for that to be ready.

**Comment by arp242** [comment]
> > So if I understand things correctly people will need a fairly recent version of illumos or it will panic? Not sure which versions of illumos distributions contain this change yet (if any?) Need to update the CI and my local test environment, as well as communicate this clearly to users.
> 
> It _might_ panic the golang application (not the operating system.) For most uses cases it probably won't. It took a fair amount of stress testing for me to trip the bug.... If it does panic, the error message points explicitly at [golang/go#54254](https://github.com/golang/go/issues/54254) which will guide users to look into an OS update.
> 
> papertigers did ping me on IRC about the need for an update for the CI OS image. Might take a little bit for that to be ready.

Ah right, I see your latest commit improved things a bit; I didn't look at the details before and just assumed it was 100% fixed on illumos side.

I think this is mostly good to merge then, right? We still have the issue of the CI not running here; I looked in that a few weeks ago, and it's not running because I changed:

    on: ['push', 'pull_request']

to:

    on:
      push:
        paths: ['**.go', 'go.mod', '.github/workflows/*']

It seems that it will only run "on" a push to this repo, but not to forks. Makes sense, but behaves different than I expected.

These "double" CI runs were causing me real grief because GitHub macOS runners are very limited, and we're spinning off 7 of them now for the various tests; I was waiting almost an hour at times for the CI to finish, so I'd rather not "just" add it back. Even with just `push` it's already pretty long waits at times.

What we need is something like:

	on: ['push_from_anywhere_including_fork']

But I couldn't really figure out how to do that :-/

Maybe I should just change it back for now...

**Comment by nshalman** [comment]
> I think this is mostly good to merge then, right? We still have the issue of the CI not running here; I looked in that a few weeks ago, and it's not running because I changed:

The behavior on symlinks you noted worries me; I'd prefer to get it fixed before landing if possible.

> The `TestWatch/watch_a_symlink_to_a_file` I added in #498 (and merged here too) seems to generate a lot of Write events:
> 
> ```
> --- FAIL: TestWatch/watch_a_symlink_to_a_file (0.76s)
>     fsnotify_test.go:214: 
>         have:
>         	WRITE                "/link"
>         	WRITE                "/link"
>         	[.. 16578 more elided ..]
>         	WRITE                "/link"
>         	WRITE                "/link"
>         want:
>         	WRITE                "/link"
> ```
> 
> The `watch_a_symlink_to_a_dir` test seems fine.

The symlink behavior quirks also remind me about #387 and all the other levels of confusion that symlinks can cause relative to fsnotify semantics...

We might want to test and document our current behavior on symlinks on Linux in a bunch of different situations and make some improved test cases around them to give me more to work with while hunting it down.

I was able to trivially make the tests pass by incorrectly changing the call to reassociate a path that had fired using lstat info, but I'm pretty sure that would fail a more robust test case of multiple write events needing to properly fire.

I'm hoping to poke around today to see if I can figure out the correct fix and will update accordingly.

**Comment by nshalman** [comment]
When I remove the calls to skip the symlink tests, sometimes I see this:
```
warning: GOPATH set to GOROOT (/home/nshalman/go) has no effect
--- FAIL: TestWatch (0.00s)
    --- FAIL: TestWatch/watch_a_symlink_to_a_dir (1.65s)
        helpers_test.go:384: this EventPort is already closed
        helpers_test.go:354: event stream was not closed after 1s
    --- FAIL: TestWatch/watch_a_symlink_to_a_file (1.70s)
        helpers_test.go:384: this EventPort is already closed
        helpers_test.go:354: event stream was not closed after 1s
```

**Comment by nshalman** [comment]
Okay, I now understand the symlink issue (passing `unix.FILE_NOFOLLOW` as part of the events and a symlink path, but the FileInfo from Stat rather than Lstat seems to basically cause the kernel to fire an event immediately and it's not wrong...)

I have a fix in place, the test skips are removed, and the tests pass!

**Comment by nshalman** [comment]
I have a reproducible test failure on illumos with go 1.19 when run as root:
```
--- FAIL: TestAdd (0.00s)
    --- FAIL: TestAdd/permission_denied (0.20s)
        fsnotify_test.go:889: error is nil
```
The test passes cleanly when run as a non-root user.
I'm guessing that on illumos the root user isn't stopped by mere permissions in that test case. 🤔

nshalman:
> I'll finish up the last few things today or tomorrow and do a new release; as far as I'm concerned it's okay to merge it after that. It doesn't need to be 100% perfect and we can create new issues and improve things before the next release. Personally I'd say that's a tad easier to keep track of things, but whatever works for you is fine with me so it's your call.

I agree that this is in a good enough state to merge.

nshalman:
@arp242 I imagine you're pretty busy, but it's been a few weeks so I figured I'd check in. Is there anything I can do to help get the next release cut so that this can be merged after that? I did fix the symlink issue, so I'm feeling very good about this code. Thanks!!

arp242:
Yeah, sorry; new job and some other issues I had to deal with 😅

I released v1.6.0 so we have a bit of time to shake out some potential issues before a new release.

arp242 (approved):

PR #374, opened 2021-07-28
Add a feature to return the directories and files that are being monitored

NitroCao:
Add a method for `Watcher` to return the directories and files that are being monitored.

`WatchList` method in `inotify.go`.

unit testing in `inotify_test.go`.


PaluMacil (approved):
Looks appropriate and tested

nathany (changes requested):
Thanks for the contribution.

Unfortunately we can't add a new `WatchList()` API that only works on Linux.

`fsnotify` is a cross-platform library. If we want to add `WatchList()`, it must also be added for macOS/BSD (kqueue) and Windows.

mattn:
@NitroCao Would you update this?

inotify.go
```diff
@@ -162,6 +162,20 @@ func (w *Watcher) Remove(name string) error {
 	return nil
 }
 
+// WatchList returns the directories and files that are being monitered.
+func (w *Watcher) WatchList() []string {
```
nathany, line 166: Be careful with adding new APIs that don't exist on other platforms. We want to keep a uniform API to the underlying systems.

@mattn

chamabreu:
Hi, is this feature comming?

windows.go
```diff
@@ -97,6 +97,19 @@ func (w *Watcher) Remove(name string) error {
 	return <-in.reply
 }
 
+// WatchList returns the directories and files that are being monitered.
+func (w *Watcher) WatchList() []string {
+	var entries []string
```
mattn, line 102:
```suggestion
	entries := make([]string, len(w.watches), len(w.watches))
```

NitroCao: Thanks for reply. updated.

mattn:
@nathany do you have opinion others?

nathany:
LGTM

mattn:
Thanks. I want to confirm rules for mainteners. Only administrators have to merge this? (I know I can merge this)

image

nathany:
Looks like there is a problem with Hound

nathany:
I removed Hound and the branch protection rule that required it.

nathany (approved):

NitroCao:
So can we merge this PR now?

mattn:
Thank you

nathany:
thanks @mattn 

I see there is a 1.5.3 tag that was created before the 1.5.2 tag for this. Will that cause problems getting this update?

mattn:
OpenBSD build was broken since before. https://github.com/fsnotify/fsnotify/pull/443

PR #379, opened 2021-08-02
Update x/sys to latest

nshalman:

Upgrades x/sys to the latest revision as of today.

This update appears to break test `(macos-latest, 1.11)`.
#371 will require an update to x/sys and exhibits the same test failure.
This PR demonstrates that the breakage is from x/sys, not the content of #371.

I don't know if it's safe to drop support for 1.11, if this represents a bug somehow introduced into x/sys/unix, or something else.

The important test failure is automated with GitHub actions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,4 @@ module github.com/fsnotify/fsnotify
 
 go 1.13
 
-require golang.org/x/sys v0.0.0-20191005200804-aed5e4c7ecf9
+require golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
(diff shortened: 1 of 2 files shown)


shogo82148:
It looks that Go 1.11 support on darwin was dropped by https://go-review.googlesource.com/c/sys/+/250437

I think we should drop support for Go 1.11 because the release policy of Go https://tip.golang.org/doc/devel/release#policy says "Each major Go release is supported until there are two newer major releases".
Latest release of Go is now 1.16 (and 1.17 will be released soon), so Go 1.14 or earlier are no longer maintained.
Users should not use them.

nshalman:
Shall I update this PR to include an update to the test matrix? Should 1.11 just be dropped entirely? Should any other releases be dropped at the same time?

shogo82148:
In my opinion, I recommend to create a pull request that drops Go 1.11.
Generally speaking, the smaller the pull request, the easier it is for the maintainers to review it.

But, I'm a newbie in fsnotify org.
@fsnotify/fsnotify Do someone have any advice?

nshalman:
> In my opinion, I recommend to create a pull request that drops Go 1.11.
> Generally speaking, the smaller the pull request, the easier it is for the maintainers to review it.

Done. See #381

Code0x58 (approved):

PR #381, opened 2021-08-04
Drop support/testing for Go 1.11 and earlier

nshalman:

Remove 1.11 (and comments about 1.10 and 1.9) from the the GitHub actions testing matrix.

See #379 where the fact that 1.11 is no longer supported becomes an issue for pulling in new features in the latest versions of x/sys/unix.

N/A


Code0x58 (approved):
seems like a good move for going forward to me

Issue #75, opened 2015-04-23
Split out packages for each OS


nathany:
Pros:
- Gives access to lower-level APIs when needed.
- Multiple high-level APIs (eg. https://github.com/rjeczalik/notify by @rjeczalik) could share some code, while having a different take on the API or supporting a subset of bindings.

Cons:
- Complicates distribution of fsnotify due to external (or internal/) packages.
- More API surface area to design (low level bindings and high level API).

We have FSEvents bindings by @samjacobson as a start:
https://github.com/go-fsnotify/fsevents

omeid:
There is no reason you can't have the combination of both. -- The `os` and `sys/GOOS` (replacing the confusing syscall) packages are a good example of how to approach a complicated set of APIs that are impossible to make strictly consistent across multiple platforms.

Or the other option is to have all the common feature set in the `fsnotify` as an interface. The package `fsnotify` then imports platform specific implementations of the aforementioned interface under the hood and allow users to cast this implementation to platform specific types if they need platform specific features.

nathany:
What I like most is that fsnotify and other libraries are free to experiment with different APIs and ways of using the underlying bindings. The lowest common denominator interface vs. emulating features in user-land vs. driver-like systems. Meanwhile the bindings themselves can be really solid and reused.

nathany:
#173 seems like a good place to start

arp242:
I think #173 is essentially the same issue, even though it's written to be about inotify, that this applies to other backends too seems like a given, even when not explicitly stated. So I will close this one.

PR #386, opened 2021-08-19
support watch files for given flags

wongoo:
add a new api  `AddWatch` which can give special flags to watch.

It just exports a new api , no logic changes.

unit test.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -88,18 +88,21 @@ func (w *Watcher) Close() error {
 	return nil
 }
 
+const agnosticEvents = unix.IN_MOVED_TO | unix.IN_MOVED_FROM |
+	unix.IN_CREATE | unix.IN_ATTRIB | unix.IN_MODIFY |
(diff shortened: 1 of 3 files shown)


nathany (changes requested):
Thanks for the contribution.

This is an interesting approach that could address #22. However, I'm not sure that we want to start exposing OS-specific APIs in `fsnotify`. The caller would need to pass in different flags for each operating system that it's compiled for.

My proposal to address things like IN_CLOSE_WRITE is #75, though open to other suggestions too.

wongoo:
@nathany  After review some issues, I get that  `fsnotify`  is designed as a cross-platform library,  which means every API must have the same behavior cross platforms. 

But from these closed issues, we can get that what users need is for a special platform. And some events only exists in one special OS platform. So I think add more capacity for users is not a bad thing, and only just need to keep the library API cross-platform itself.

How about change name of the new API to `AddOSWatch` to tell users the watch is only for special platform ?

nathany:
That's a reasonable idea, but let's move this to a discussion instead of a pull request. #392

Issue #387, opened 2021-08-19
Confusion about Remove when combined with Add vs AddRaw


nshalman:
### Please describe the issue that occurred.

While porting #371 to support AddRaw from #289 I run into the following conundrum:

If I AddRaw a symlink, I can Remove that symlink from being watched just fine.
If I Add a symlink, I cannot Remove the symlink directly, but would have to resolve it first.

What should fsnotify do?

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

See the test case added in #388

nathany:
It does seem like a bug if Remove doesn't resolve symlinks the way Add does. Is that across every OS?

I'm not sure if we should go as far as `Remove` and `RemoveRaw` APIs.

To be honest, for the simplicity of the API, I'm starting to wonder if we should've made a breaking change (major version) that pushes the responsibility for `EvalSymlinks` up to the caller. 🤔 Probably not, but I had that thought.

@Code0x58 What do you think?

Code0x58:
I added a comment on https://github.com/fsnotify/fsnotify/pull/388#discussion_r693552563. I'm not sure that a major change is worth it at the moment, as removing `Add`/`Remove` or replacing them with `AddRaw`/(presumably)`RemoveRaw` - there's still hope that `Add`/`Remove` could be incrementally improved to the point that they don't come with old caveats, so could be a nice transparent upgrade in the future.

nshalman:
Fixed by #394

PR #382, opened 2021-08-04
Lint

Code0x58:
* remove linting issues
* require future PRs to not introduce any linting issues
* fix tests on Windows which were denied access when trying to write to a file opened for only reading
* remove testRename as an external call isn't necessary

_**edit:** crudely fixing the lint issues shows a lot of returned errors which is making this PR grow. The lint golangci-lint job doesn't show everything at once so can be a game of whack-a-mole. Changing PR to draft until everything is sorted out_

PR #388, opened 2021-08-19
Test Symlink Add and Remove

nshalman:

Adds a test case to highlight the confusion about symlinks in #387 

Please look at the test failures and the code and provide guidance.

```go test -v -run TestSymlinkWatchRemoval```

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -1076,6 +1076,53 @@ func TestSymlinkNotResolved(t *testing.T) {
 
 }
 
+func TestSymlinkWatchRemoval(t *testing.T) {
+	testDir := tempMkdir(t)
+	file1 := filepath.Join(testDir, "file1")
+	link := filepath.Join(testDir, "link")
+
+	f1, err := os.Create(file1)
+	if err != nil {
+		t.Fatalf("Failed to create file1: %s", err)
+	}
+	defer f1.Close()
+
+	// symlink works for Windows too
+	if err := os.Symlink(file1, link); err != nil {
+		t.Fatalf("Failed to create symlink: %s", err)
+	}
+
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("Failed to create watcher")
+	}
+
+	// if we AddRaw a symlink
+	err = w.AddRaw(link)
+	if err != nil {
+		t.Fatalf("Failed to add raw symlink: %s", err)
+	}
+
+	// that path name will be watched and can be unwatched
+	err = w.Remove(link)
+	if err != nil {
+		t.Fatalf("Failed to remove link: %s", err)
+	}
+
+	// but if we Add a symlink
+	err = w.Add(link)
+	if err != nil {
+		t.Fatalf("Failed to add link: %s", err)
+	}
+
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: integration_test.go [hunk 1]
```diff
@@ -1076,6 +1076,53 @@ func TestSymlinkNotResolved(t *testing.T) {
...
+	err = w.Add(link)
+	if err != nil {
+		t.Fatalf("Failed to add link: %s", err)
+	}
+
+	// what should we do if a user tries to Remove it?
+	err = w.Remove(link)
+	if err != nil {
+		t.Fatalf("Failed to remove link: %s", err)
+	}
```
nathany, line 1122: are you saying that Remove doesn't follow symlinks? (effectively RemoveRaw)

@Code0x58 What do you think?

nshalman: The CI tests confirm the issue for Mac and Linux. I had the same problem with my solaris/illumos port.
Windows doesn't seem to care, but Windows also doesn't really have symlinks.

My immediate thought was that a Remove that fails could do an extra check to see if the requested path to remove is a symlink and do the same resolution and try again, but that should probably be implemented in fsnotify.go across all platforms rather than individually in each os-specific implementation.

Definitely curious to know what @Code0x58 thinks we should do.

Code0x58: Good spot and test!

For now, I think renaming `Remove` to `RemoveRaw` and making the Windows implementation consistent is the way to go, as well as adding a naive solution similar to `Add` (i.e. `Remove` tries to deference the symlinks). The naive `Add` and `Remove` would come with caveats to document - I'd probably recommend not using `Add`/`Remove` at all unless you know their caveats work in your use case. I'm not sure it would fix any existing users for users, aside from inconsistency with Windows/other implementations.

With a portable low level interface, and documentation on existing issues, the library would be in a better position for users, and may attract contributions towards documenting or implementing common patterns, and maybe even a generic high level interface. A vague comment on a high level interface is [here](https://github.com/fsnotify/fsnotify/issues/199#issuecomment-480277398), but it could probably use a lot of attention to avoid leading users in to problematic cases (with races/consistency). If the retry approach would stop a potential issue, and not introduce any more, then I'd say it's a good thing to do so that the nativity/caveats of `Add`/`Remove` is incrementally reduced.

_ramble:_ I think the codebase would need a fair bit of fixing and ramping up in testing to offer a properly dependable cross platform solution for non-trivial watching (trivial such as watching a file in a directory which won't be moved or deleted). I say this following experience trying to fix the hidden issues found by linting (#382 - usually where errors aren't checked in tests, or test objects are used in goroutines which shouldn't - so may not be working at all). That could be a quite a bit of a task taking a lot of attention, but documenting and testing those safe patterns would be good and attract a community. I imagine a fair few people would like a robust pattern for matching a single config file which may be a file, or a symlink to a file such as when you are making a program which may run in or out of a container, and the config may be a projection mount within a container (e.g. when using Docker, Kubernetes, etc.) - that was what inspired my original PR back in 2019.

nshalman:
With the revert of #289 in #394, this is not currently needed.
The test case can be incorporated into the next iteration of better symlink handling.

Issue #393, opened 2021-08-24
Error change in 1.5.0 when adding a file that doesn't exist


oschwald:
Previously, if `Add` was called on a file that doesn't exist, a `syscall.ENOENT` error would be returned. We used this to handle the case where the file may or may not exists, e.g.:

```go
	err = watcher.Add(lockFile)
	if err != nil {
		if errors.Is(err, syscall.ENOENT) {
			...
		}
		...
	}
```

As of https://github.com/fsnotify/fsnotify/pull/289, this error is now returned:

https://github.com/fsnotify/fsnotify/blob/e2e95171bc7c7402864e613c8b00a768fbcc2380/fsnotify.go#L76

Given that this is a new error that doesn't wrap the original error, the only option to check if the cause is that the file doesn't exist is to try to match against the error message. This seems fragile and is an API change.

We rely on the `syscall.ENOENT` error from the `Add` as checking whether the file exists before the `Add` would just introduce a race condition, where the file could be remove after the check but before the `Add`.

Note: we have worked around this by switching to `AddRaw`, but the behavior on `Add` could be improved.

nathany:
v1.5.1 reverts the changes from 1.5.0, but lets leave this open to review.

arp242:
Closing as duplicate of #144

All that was needed was changing the `%s` to a `%w` really, but no point in keeping this issue open and Add() error are covered in the other issue anyway.

Issue #390, opened 2021-08-23
Symlink regression in v1.5.0


bep:
See https://github.com/gohugoio/hugo/issues/8920

On Linux.
 
/cc @pamubay

nathany:
Thanks for the report.

@Code0x58 Are you willing to look into this?

nathany:
There appear to be quite a number of regressions in v1.5.0 in addition to this one: #393 #389.

Thinking we may need to rollback the changes for now -- so we can take our time on giving it another try.

/cc @Code0x58 @nshalman

nshalman:
> Thinking we may need to rollback the changes for now -- so we can take our time on giving it another try.
I support that.

nathany:
v1.5.1 reverts the changes from 1.5.0, but lets leave this open to review.

mterwill:
👋 any plans to revisit adding support for watching symlinks without following? Thanks!

arp242:
> any plans to revisit adding support for watching symlinks without following? Thanks!

This can't be easily changed without breaking compatibility. The current situation is unfortunate because the behaviour is inconsistent on different platforms, but clearly people rely on it so it can't "just" be changed. The current API also doesn't give us much room for adding options to Add() or some such, so idk...

Either way, this is a duplicate of #227, so closing it.

Issue #389, opened 2021-08-23
fsnotify v1.5.0 and GOOS=freebsd does not build


xyproto:
`GOOS=freebsd go` build works with Go 1.17 and fsnotify v1.4.9, but stops working with fsnotify v1.5.0.

The error message is:

```
vendor/github.com/fsnotify/fsnotify/kqueue.go:193:43: undefined: unix.O_SYMLINK
```

nathany:
Thanks for the report.

Confirmed that changes in 1.5.0 via #289 began using `unix.O_SYMLINK` which is defined for Darwin (macOS) [zerrors_darwin_amd64.go](https://cs.opensource.google/go/x/sys/+/master:unix/zerrors_darwin_amd64.go;l=1138) but not for FreeBSD or other BSDs.

/cc @Code0x58

nathany:
Of note, we don't have FreeBSD as a GitHub runner, because they only supports Linux, Windows, and macOS. They do support custom runners as one possibility. Or we may be able to use the the same trick that #371 is using for Solaris?
https://github.com/papertigers/illumos-vm

I don't currently have a FreeBSD environment set up -- I was using Vagrant before, but I haven't worked with it for a while.
https://github.com/nathany/vagrant-gopher

nshalman:
> Of note, we don't have FreeBSD as a GitHub runner, because they only supports Linux, Windows, and macOS. They do support custom runners as one possibility. Or we may be able to use the the same trick that #371 is using for Solaris?

See https://github.com/vmactions/freebsd-vm

nathany:
v1.5.1 reverts the changes from 1.5.0, but lets leave this open to review.

nathany:
I do not have BSD to test on. Please report back whether 1.5.1 resolved the issue.

r-darwish:
Ran `go test` on FreeBSD 13 with Go 1.17 on the main branch. All tests passed

nshalman:
@r-darwish Any chance you could help me write a github action that uses https://github.com/vmactions/freebsd-vm similar to the one I'm working on for illumos (see https://github.com/nshalman/fsnotify/commit/96066fd2575e3fa1fc8510d4cd0cbcb021337a05) to run the tests on FreeBSD?

It would be great to be able to catch these things in the future, even if we have to trigger the tests manually because they're too slow to run on every push/pr.

r-darwish:
https://github.com/fsnotify/fsnotify/pull/419

[...]

PR #394, opened 2021-08-24
Revert Add AddRaw to not follow symlinks

nathany:

Revert #289 to resolve regressions in #390 #393 #389.

We can give it another go. @Code0x58.

`go test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,18 +72,10 @@ func main() {
 		}
 	}()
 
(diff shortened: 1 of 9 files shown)


nshalman (approved):
This revert is fine. We should probably also update go.mod to retract v1.5.0

See https://golangtutorial.dev/tips/retract-go-module-versions/

PR #400, opened 2021-10-02
Remove `retract` directive

y-yagi:

Remove `retract` directive.

The `retract` directive only works Go >= 1.16. The versions under that can't recognize the directive and cause an error. (e.g. https://github.com/fsnotify/fsnotify/runs/3414923756)

So removed the directive to work builds.

Please see CI passes.

No need.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,5 +3,3 @@ module github.com/fsnotify/fsnotify
 go 1.13
 
 require golang.org/x/sys v0.0.0-20210630005230-0f9fa26af87c
-
-retract v1.5.0


shogo82148:
The latest version of Go is 1.17.5, and supported versions of Go are 1.17.5 and 1.16.12.
https://go.dev/doc/devel/release#policy

Go 1.15 and older are already unsupported.
Should we drop support of them?

nathany:
@y-yagi Thanks for bringing this up, but I agree with @shogo82148 -- we should only need to support just the last 2 versions of Go.

Also, fsnotify 1.5.0 had a number of regressions, so I think it's important to prevent people from using it. 
#390 #393 #389.

PR #406, opened 2021-12-10
Don't set `poller.fd` twice in `newFdPoller`

tklauser:
In `newEmptyPoller` when creating the `poller` using `newEmptyPoller(fd)`, the
`fd` field is already set. There is no need to set it again.

### DIFF
--- a/inotify_poller.go
+++ b/inotify_poller.go
@@ -38,7 +38,6 @@ func newFdPoller(fd int) (*fdPoller, error) {
 			poller.close()
 		}
 	}()
-	poller.fd = fd
 
 	// Create epoll fd
 	poller.epfd, errno = unix.EpollCreate1(unix.EPOLL_CLOEXEC)


nshalman:
I thought the tests would run when I did that... 😕

nshalman:
@tklauser could you rebase your commit and force push to replace my merge commit to see if that will trigger the tests?

inotify_poller.go
```diff
@@ -38,7 +38,6 @@ func newFdPoller(fd int) (*fdPoller, error) {
 			poller.close()
 		}
 	}()
-	poller.fd = fd
```
nshalman, line 41: Just to be super explicit about this, line 35 calls `emptyPoller(fd)` and on line 24 `poller.fd = fd`

nshalman (approved):
This makes sense to me. Being extra cautious and requesting @shogo82148 to confirm too.

nshalman:
Having stared at this enough times and seeing the tests pass, I'm satisfied.

Issue #412, opened 2022-01-14
go vet: call to (*T).Fatalf from a non-test goroutine


nathany:
**Describe the bug**

go vet reports the following issue in our tests

./inotify_poller_test.go:182:5: call to (*T).Fatalf from a non-test goroutine
./inotify_test.go:341:4: call to (*T).Fatalf from a non-test goroutine
./integration_test.go:80:4: call to (*T).Fatalf from a non-test goroutine
./integration_test.go:843:4: call to (*T).Fatalf from a non-test goroutine

**To Reproduce**

Run `go vet` locally.

**Expected behaviour**

Errors need to be propagated back to the main goroutine where t.Fatalf can be called. Code like this needs to be updated:

```go
go func() {
	for err := range watcher.Errors {
		t.Fatalf("error received: %s", err)
	}
}()
```

This needs to be resolved for CI to pass -- the first step before anything else.

**Which operating system and version are you using?**

all operating systems

PR #414, opened 2022-01-14
pass testing.T explicitly to the anonymous func

hidetatz:

Passes testing.T explicitly to the anonymous function run as a goroutine. By this change, `go vet` succeeds.
Fixes https://github.com/fsnotify/fsnotify/issues/412

Read https://github.com/fsnotify/fsnotify/issues/412 to understand the problem and make sure testing.T is correctly used.

Run `go vet` locally.

### DIFF
--- a/inotify_poller_test.go
+++ b/inotify_poller_test.go
@@ -174,19 +174,19 @@ func TestPollerConcurrent(t *testing.T) {
 	oks := make(chan bool)
 	live := make(chan bool)
 	defer close(live)
-	go func() {
+	go func(tb testing.TB) {
 		defer close(oks)
 		for {
 			ok, err := poller.wait()
 			if err != nil {
-				t.Fatalf("poller failed: %v", err)
+				tb.Fatalf("poller failed: %v", err)
 			}
 			oks <- ok
 			if !<-live {
 				return
 			}
 		}
-	}()
+	}(t)
 
 	// Try a write
 	select {
--- a/inotify_test.go
+++ b/inotify_test.go
@@ -336,11 +336,11 @@ func TestInotifyInnerMapLength(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Failed to add testFile: %v", err)
 	}
-	go func() {
+	go func(tb testing.TB) {
 		for err := range w.Errors {
-			t.Fatalf("error received: %s", err)
+			tb.Fatalf("error received: %s", err)
 		}
-	}()
+	}(t)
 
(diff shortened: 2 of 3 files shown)


shogo82148 (changes requested):
Thanks for your contribution, but it just deceives `go vet` and is not an essential solution.
We need to wait for all the goroutines to finish before calling `(*T).Fatalf`.

nathany:
I did see some example solutions when I searched the Internet for the vet error... there's a better way to do it, I'm certain.

Little point in running go vet if we're going to trick it.

hidetatz:
Does it make sense to use errgroup? I'm writing patch like this:

```diff
diff --git a/inotify_poller_test.go b/inotify_poller_test.go
index 0570057..80a2bb5 100644
--- a/inotify_poller_test.go
+++ b/inotify_poller_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"golang.org/x/sync/errgroup"
 	"golang.org/x/sys/unix"
 )
 
@@ -174,19 +175,23 @@ func TestPollerConcurrent(t *testing.T) {
 	oks := make(chan bool)
 	live := make(chan bool)
 	defer close(live)
-	go func(tb testing.TB) {
+	eg := errgroup.Group{}
+	eg.Go(func() error {
 		defer close(oks)
 		for {
 			ok, err := poller.wait()
 			if err != nil {
-				tb.Fatalf("poller failed: %v", err)
+				return err
 			}
 			oks <- ok
 			if !<-live {
-				return
+				return nil
 			}
 		}
-	}(t)
+	})
+	if err := eg.Wait(); err != nil {
+		t.Fatalf("poller failed: %v", err)
+	}
 
 	// Try a write
 	select {

```

nathany:
That looks like a very nice solution

It does mean adding another dependency, but only for tests. 🤔

hidetatz:
Another way is using sync.WaitGroup instead of errgroup if introducing another dependency is unacceptable, but using errgroup is clearer way for me.
OK, I'll open another PR. 

> There are multiple platforms where the vet error occurs.

Got it, I think something like `GOOS=windows GOARCH=amd64 go vet ./...` will work.

nathany:
👍🏼 I'll let you decide -- if you can make sync.WaitGroup work nicely (maybe adding a helper function or what not), that may be preferred.

Kind've need a way to trigger the error case to make sure it all works, because normally Fatalf shouldn't be called. :-)

PR #415, opened 2022-01-14
Re-enable tests for PRs

nshalman:
I imagine that disabling testing for PRs may have been needed to unbreak things while cleaning up the older releases in the test matrix.
It would be good to re-enable tests on PRs.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,7 @@
 name: test
-on: [push]
+on:
+  push:
+  pull_request:
 jobs:
   test:
     strategy:


nshalman:
Ah, lint is unhappy. I think there are other PRs that relate.

nshalman:
@shogo82148 Do you think it will be easy to safely clean up the lint errors? Should we re-enable tests and for now just ensure that no NEW lint failures creep in?

nathany:
I don't think this is necessary. Opening a pull request is also a push. If you look through the checks, you can see twice as many tests running as needed. Trying to save some cycles.

If you want to take a closer look at GitHub's docs for Actions, I only looked at the very first example I saw, which was on push only.

As far as linting, #382 attempts to do that, but there's a lot there.

Maybe it would make sense to disable the linting until it's fixed, or even selectively disable certain errors that are "TODOs" to fix. At least then the build would pass?

shogo82148:
I opened https://github.com/fsnotify/fsnotify/pull/416 for fixing the lint.

> If you look through the checks, you can see twice as many tests runs as needed. Trying to save some cycles.

That's right.
But note that the commit hashes to be checked out are different for `push` and `pull_request` events.
The commit hash is `refs/pull/:prNumber/merge` on `pull_request` events, however it is "updated ref" (typically same as `refs/pull/:prNumber/head`) on `push` events.

- https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#push
- https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request

Generally speaking, `refs/pull/:prNumber/merge` and `refs/pull/:prNumber/head` have different contents.
Because someone might commit to the base branch before opening the pull request.

nshalman:
The `pull_request` event also annotates the PR more clearly with the success or failure which I find to be helpful when reviewing PRs, as well as being usable to block merging when tests are failing. I'll take a look at #416.

nathany:
Do you need both push and pull_request though?

I'd prefer just one or the other - either one.

Btw, the `main` branch has the GitHub setting enabled "Require branches to be up to date before merging"

nathany:
We can also add `test (macos-latest, 1.17)`, etc. to the list of status checks that are required in the branch settings. Does that annotates the PR more clearly?

nathany (approved):
I don't like wasting cycles, but if you feel it's helpful, then 👌🏼

nshalman:
> I don't like wasting cycles, but if you feel it's helpful, then 👌🏼

Worse than wasting computer cycles is wasting maintainer cycles looking for things that computers could have caught.

PR #418, opened 2022-01-16
Fix 288: close handle when remWatch open in getIno

mattn:

Closing #288

Commits are contained in this PR too.

No


mattn:
@shogo82148 Could you please approve this? #288 have to be fixed since this is handle leak. On Windows, the directory can not be possible to delete manually if any handles are opend for the directory.

nshalman:
Is there a reason to only run this test on Windows?

mattn:
This is a bug occured only on Windows.

windows_test.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	watcher := newWatcher(t)
+	defer watcher.Close()
+
+	addWatch(t, watcher, testDir)
+	if err := watcher.Remove(testDir); err != nil {
+		t.Fatalf("Could not remove the watch: %v\n", err)
+	}
+	if err := watcher.remWatch(testDir); err == nil {
```
nshalman, line 40: If this line was also` watcher.Remove` I think this should be safe to put back in as an integration test across all OSes. It works correctly on my as-yet-not-merged illumos tree, and I think we should do our best that most tests should be run across all platforms rather than specific to a single one...

nshalman (changes requested):
Unless I'm really misunderstanding something, I think this can be made into an integration test which is generally preferable. Comment inline explaining how I think it could be done.

windows_test.go
mattn: Okay, added

nshalman: Uhhh. Those test failures are confusing me. I wonder if the official API is to allow repeated removal and just ignore it if it's not watched? Let me do some digging. Thank you for being responsive! Sorry for the delay, but I think we can get this right.

integration_test.go
```diff
@@ -873,6 +874,40 @@ func TestRemovalOfWatch(t *testing.T) {
...
+	}
+
+	watcher := newWatcher(t)
+	defer watcher.Close()
+
+	addWatch(t, watcher, testDir)
+
+	var err error
+	if err = watcher.Remove(testDir); err != nil {
+		t.Fatalf("Could not remove the watch: %v\n", err)
```
nshalman, line 901: I don't think we want/need the `\n`

integration_test.go
```diff
@@ -873,6 +874,40 @@ func TestRemovalOfWatch(t *testing.T) {
...
+	}
+	if err = watcher.Remove(testDir); err == nil {
+		t.Fatalf("Shold: %v\n", err)
```
nshalman, line 904: Perhaps `t.Fatalf("Removal of already removed watcher should have failed")`

integration_test.go
```diff
@@ -873,6 +874,40 @@ func TestRemovalOfWatch(t *testing.T) {
...
+	}
+	if !strings.HasPrefix(err.Error(), "can't remove non-existent ") {
+		t.Fatalf("Should be fail with removing non-existence watcher: %v\n", err)
```
nshalman, line 907: Perhaps `t.Fatalf("Expected error to start with \"can't remove non-existent \": %v", err)`

nshalman (commented):
I think we're almost there. Unless there's a really good reason to have a duplicate test on Windows I think we just need the fixed up integration test and the fix for Windows.

Can you confirm that the integration test fails on Windows without the fix?

integration_test.go
mattn: Yes, there are many \n in `*_test.go`. If you want to fix, I'll fix this in new PR.

integration_test.go
mattn: fixed

integration_test.go
mattn: Fixed

integration_test.go
mattn: For the comment about `remWatch`, I'm thinking this test is required since TestRemovalOfWatchTwice does not hit this bug strictly.

nshalman: Ah, then ignore my feedback on that. Consistent is fine. 😄

nshalman:
I don't think this test catches the corresponding error. I tried pushing a branch with the test but not the fix and the tests still pass: https://github.com/nshalman/fsnotify/actions/runs/1709124231

I think we should look more closely at https://gist.github.com/timshannon/603f92824c5294269797 and https://github.com/fsnotify/fsnotify/issues/42

mattn:
Thank you. I merged your patch.

shogo82148 (approved):
LGTM, thanks!

nshalman:
Sorry, I don't think I was clear. My test is not diagnostically useful in its current form. It passes even without the proposed fix. :(

For a regression test to be useful, it should fail if run before the fix is applied. I could use some help writing an effective regression test.

mattn:
I'll look in later.

arp242:
I just merged this as it's been well over a year; if more work needs to be done then this can be a new PR.

PR #419, opened 2022-01-17
Add FreeBSD testing in Github Actions (fix #389)

### LINKED ISSUE #389: fsnotify v1.5.0 and GOOS=freebsd does not build

r-darwish:
Fix #389

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,7 @@ jobs:
 
   testFreeBSD:
     runs-on: macos-10.15
+    name: test (freebsd, 1.17)
     steps:
       - uses: actions/checkout@v2
       - name: test (freebsd, 1.17)
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,12 +28,26 @@ jobs:
         run: |
           go test --race ./...
 
+  testFreeBSD:
+    runs-on: macos-10.15
+    name: Test FreeBSD
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test in FreeBSD
+        id: test
+        uses: vmactions/freebsd-vm@v0.1.5
+        with:
+          usesh: true
+          prepare: pkg install -y go
+          run: |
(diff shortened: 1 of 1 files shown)


nshalman:
Thank you!!
It looks like that test [took about 5 minutes to run](https://github.com/r-darwish/fsnotify/runs/4844833573?check_suite_focus=true). My only question is whether we want that on every push or if we want to just give maintainers the option to run them manually.
Either way, this is fantastic to have and I look forward to some version of it landing soon!

r-darwish:
Yeah, the reason for being slow is that it downloads a full VM image, extract it and boots it up before doing anything. I think 5 minutes are bearable, though

nathany (approved):

nshalman:
I wonder if there's a clever way to make it somehow depend on the cross-compilation target so that if it wouldn't build we won't waste time spinning up the VM for testing...

nathany:
Well, if we do `needs: build`, it will run sequentially instead of in parallel...

nshalman:
The builds are very fast. I see the argument on the other side though.

nathany:
That the VM build starts later? Maybe more waiting

nathany:
How can we get builds to run on external pull requests like this?

We don't have any secrets to worry about leaking.

nshalman:
> How can we get builds to run on external pull requests like this?
> 
> We don't have any secrets to worry about leaking.

😄  #415 should do it. Alternately we could favor `pull_request` over `push` though I do still prefer both.

nshalman:
Given @nathany's approval I think we should merge this. If the build times become onerous we can revisit moving them out to the forthcoming slow-tests that I'm planning for illumos. Better to err on the side of testing.

PR #420, opened 2022-01-17
Run cross-compilation builds on every push

nshalman:

This creates a Github action to run cross-compilation
builds for all supported targets (or at least that's the idea)

This would have caught #389 and should
catch build failures for supported targets.

The output of the builds

N/A

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,30 @@
+name: build
+on: push
+jobs:
+  cross-compile-supported:
+    strategy:
+      fail-fast: false
+      matrix:
+        goos:
+          - android
+          - darwin
+          - freebsd
+          - ios
+          - linux
+          - windows
+        goarch:
+          - amd64
+          - arm64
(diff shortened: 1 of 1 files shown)


nshalman:
Builds ran in my fork: https://github.com/nshalman/fsnotify/actions/runs/1709688309

nshalman:
Even on my illumos system I could have caught that:
```
~/fsnotify $ git checkout v1.5.0 ; GOOS=freebsd GOARCH=amd64 go build
Note: switching to 'v1.5.0'.
<snip>
HEAD is now at 08848a0 v1.5.0 preparation (#380)
# github.com/fsnotify/fsnotify
./kqueue.go:193:43: undefined: unix.O_SYMLINK
```

nshalman:
I think this PR is a good example of how I intend to maintain quality in this project across all supported platforms. I hope you agree, @nathany.

nathany (approved):
very good idea :shipit:

PR #422, opened 2022-01-20
Integration Tests: change 1ms sleeps to 50ms

nshalman:

When testing the FEN code repeatedly and under
system load I can sometimes get these tests to fail
with a missed WRITE event. With these longer sleeps
I have been unable to produce any test failures.

This is unashamedly in support of #12 / #371

Additionally, this PR gives slightly more friendly names to the sleep durations based on their purposes.

Is there any good reason not to do this? It seems reasonable to me and
it prevents spurious test failures.

N/A

### DIFF
--- a/integration_test.go
+++ b/integration_test.go
@@ -135,7 +135,7 @@ func TestFsnotifyMultipleOperations(t *testing.T) {
 	}
 	f.Sync()
 
-	time.Sleep(time.Millisecond)
+	time.Sleep(50 * time.Millisecond)
 	f.WriteString("data")
 	f.Sync()
 	f.Close()
@@ -248,7 +248,7 @@ func TestFsnotifyMultipleCreates(t *testing.T) {
 	}
 	f.Sync()
 
-	time.Sleep(time.Millisecond)
+	time.Sleep(50 * time.Millisecond)
 	f.WriteString("data")
 	f.Sync()
 	f.Close()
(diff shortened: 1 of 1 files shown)


integration_test.go
```diff
@@ -135,7 +135,7 @@ func TestFsnotifyMultipleOperations(t *testing.T) {
 	}
 	f.Sync()
 
-	time.Sleep(time.Millisecond)
+	time.Sleep(50 * time.Millisecond)
```
nathany, line 138: maybe it would be worth defining a `const`?

nathany (approved):

integration_test.go
nshalman: Naming is hard. Any suggestions for what to name it? I'm happy to make that change.
I was searching incorrectly for `static` for a different line in the FEN code. What I wanted is `const`, so this was also serendipitous.

nathany: Naming is hard. Brainstorming 🌩️ 

It's a `sleepDuration` ... but to be more specific, it's the time to wait for events.

It is a `timeout` of sorts, because we assume it failed after this time elapses.

Uhm... I don't know. What do you think?

nshalman: `shortNap` ?

nathany: 🤷🏼

nshalman: We also have a 500ms sleep which is for waiting for the events to finish rolling in that could also be refactored.
call that `longNap` while I'm in there?

nshalman: `eventSeparator` (50ms) and `waitForEvents` (500ms) or something like that could be a little more descriptive?

nshalman: I've implemented those last two, but they are easy string replacements with something better.

nathany (approved):

PR #427, opened 2022-01-28
Add raw, os-dependent op bitmask to Event

roobre:

Adds the raw, unprocessed event bitmask to the `Event` struct.

I have seen a couple of PRs closed when adding events that were not cross platform, referring to a [future lower-level API](https://github.com/fsnotify/fsnotify/issues/173), which in no doubt is the best way forward. However, building that API seems to be a long-term effort.

In my view there are use cases that would benefit from using fsnotify in a way that is not completely cross platform, and willing to forfeit validation on its side For example:
- Small projects which are targeted to run in a particular OS
- Larger projects which are willing to deal with platform-dependent code in their side

With this use cases in mind, I think there could be some value in having the raw, os-dependent op bitmask reported in the event, in a non-intrusive way. Callers should be aware that this bitmask is heavily OS-dependent, and act accordingly in their side instead.

The whole PR should fit in the viewport :)

I don't think there are any tests needed, but if desired, I can add a few integration tests build-tagged for unix and asserting the reported `Event.RawOp`.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -16,8 +16,9 @@ import (
 
 // Event represents a single file system notification.
 type Event struct {
-	Name string // Relative path to the file or directory.
-	Op   Op     // File operation that triggered the event.
+	Name  string // Relative path to the file or directory.
+	Op    Op     // File operation that triggered the event.
+	RawOp uint32 // Unprocessed event bitmask as returned by the OS.
 }
 
 // Op describes a set of file operations.
--- a/inotify.go
+++ b/inotify.go
@@ -318,7 +318,10 @@ func (e *Event) ignoreLinux(mask uint32) bool {
 
 // newEvent returns an platform-independent Event based on an inotify mask.
 func newEvent(name string, mask uint32) Event {
-	e := Event{Name: name}
+	e := Event{
+		Name:  name,
+		RawOp: mask,
+	}
 	if mask&unix.IN_CREATE == unix.IN_CREATE || mask&unix.IN_MOVED_TO == unix.IN_MOVED_TO {
 		e.Op |= Create
 	}


marklr:
Why not add the raw op to kqueue events too?

roobre:
> Why not add the raw op to kqueue events too?

@marklr I'm not entirely sure what you mean but I'm open to making changes if the maintainers or the community think it would be better :)

nshalman (changes requested):
First of all, thank you for continuing to pursue making fsnotify better.

I think the other commenter was trying to say something along the lines of "If you're implementing the RawOp for inotify (Linux), why not also do the work to implement it for kqueue(Mac OS and FreeBSD)"

Certainly I'm not comfortable merging this sort of thing without test code that verifies that it works so that we ensure we don't break it going forward.

I too would prefer that if we're going to try to implement it that we implement it for all platforms. I have a reasonable sense of how I would do it for my nearly-ready-to-land illumos/Solaris FEN work (#371) (though the more I think about it, the weirder things get; I'm not sure that RawOp would totally make sense on that platform...)

So, If nothing else we would need some inotify-specific tests to prevent breaking it in the future, and an integration test that could perhaps do some special casing by GOOS until it's fully implemented for all platforms.

Also, if you have a moment to link to the other closed issues you had in mind, that might help me understand what the uses for RawOp are.

roobre:
Hey @nshalman, thanks a lot for your thoughts! I'll try to address your points when I have some time during the weekend. I have an osx machine so I might be able to try out how this could work for kqueue.

Regarding the tests, I'm more than happy to add them, but I do wonder what is the scope you have in mind for those exactly. As the feature is now, it essentially copies whatever the OS is reporting. If I make some integration tests on that, wouldn't I be testing the underlying implementation (e.g. `inotify`) and not `fsnotify` itself? If you can clarify a bit what would you like tests to cover exactly, I'll happily add some :)

nshalman:
The more I think about this the less of a good idea it seems and the more it feels like this is related to #173 (as you already linked.)
Any software that takes advantage of the raw events for a specific platform e.g. Linux is at risk of becoming Linux-only (possibly by accident!) in spite of having used fsnotify.

A factored out inotify package could/should be free to expose those raw events to pieces of software that are intending to be Linux-only.

So let's think a little more. What kind of information would be useful from the RawOp and could we perhaps just extend the fsnotify to get that information on all platforms? Examples would be very helpful.

roobre:
Hey @nshalman,

Thanks a lot for your thoughts. In thiis particular case, the need I have is for purely the `OPEN` event, which seems to be linux-specific.

> Any software that takes advantage of the raw events for a specific platform e.g. Linux is at risk of becoming Linux-only (possibly by accident!) in spite of having used fsnotify.

For my personal use case, I was willing to take this tradeoff. I though that, even without using its cross-platform features, `fsnotify` could still provide a nice and user-friendly API when compared to lower-level primitives. Is this a use case you don't agree on?

arp242:
I agree that adding a field and setting it for just one watcher type/platform is not ideal; a huge reason for using this in the first place is for its cross-platform support, and this somewhat "silently" drops it. A lot of applications support FreeBSD, macOS, and other platforms but they're often not as well tested, or not tested at all, and these kind of runtime incompatibilities are tricky to catch in the first place.

I do appreciate there are real-world pragmatic concerns and I think it's a valid use case worth solving, but I think this has the potential to cause a lot of confusion and breakage down the line.

A nice solution might be making `Event` an interface and returning `EventInotify` as an instance of that (which has `RawOp()`), but that's not a compatible change :-/

I'm not entirely sure what the best path forward is here, but I don't think that this PR is it. Doing the previously mentioned "low-level inotify package" (#173) probably remains the best option.

arp242:
I'll close this PR for now, but feel free to get back with other potential solutions.

Issue #426, opened 2022-01-26
Could we please get back the master-branch on this repository?


SimonSimCity:
This request not about anything else but to get our build scripts to run again.

In our scenario, we use Yocto and its meta-layer `meta-virtualization`, which includes a script of how to generate this package. Until 12 days ago, the repository from where to pull the codebase was noted as `git://github.com/fsnotify/fsnotify.git;branch=master;protocol=https`. Since the branch `master` was removed from this repository, all previous states of the repository cannot be used to build a valid Linux distribution without first having to change the branch name.

For us this now creates a really big challenge between not being able to build. Solutions we see on our side are getting everything up to date, which would require a ton of time and testing; or to just fork the repo and having to maintain it ourselves from the state we're now on - more maintenance.

I guess many other people are facing the same problem, where I just want to ask - what's wrong with keeping it called `master` vs breaking almost any build script which relies on the existence of the branch?

sagikazarmark:
Renaming a master branch in Go libraries is especially problematic. Although I understand that this is a sensitive issue, I believe we should prioritize the backward compatibility, so +1 from me restoring the master branch with a big fat deprecatio n alert on the readme that it's no longer the default branch. (Alternatively, we could sync commits from main, but I don't think that's necessary)

TommyDuc:
I've just add the same issue and as OP stated. There is no easy way for us to go around this issue, especially in the context of automating the build of another client's repository. The 'master' branch not being the default branch is OK for us but it is way too early to delete it entirely.

kolyshkin:
What needs to be done locally (in a clone of this repo) to rename master to main is:
```bash
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```
and, if you want to remove the leftover `origin/master` branch (and all the other origin/* branches that no longer exist in `origin`):
```bash
git remote prune origin
```

This is a one time thing.

HenkKalkwater:
In my opinion, if you want to continue working on a `main` branch that is fine. But please leave the old master branch to prevent builds from suddenly breaking in the future. You don't have to push new commits to this branch. If the version of the project gets updated in those build scripts, changing the branch name is not a problem.

Please don't rewrite or erase git history of important branches.

sagikazarmark:
> > Renaming a master branch in Go libraries is especially problematic.
> 
> In what way? I've not encountered any problems and I've renamed a bunch.

[...]

PR #429, opened 2022-02-16
[DO NOT MERGE] README: warning about obsoleted master branch

kolyshkin:

This is not a pull request, but rather a way to discuss the wording in README.md for the
proposed `master` branch, needed to alleviate #426.

Once the wording is agreed upon, I will create a new `master` branch in this repo.

#426

n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,130 +1,40 @@
-# File system notifications for Go
+# WARNING
 
-[![GoDoc](https://godoc.org/github.com/fsnotify/fsnotify?status.svg)](https://godoc.org/github.com/fsnotify/fsnotify) [![Go Report Card](https://goreportcard.com/badge/github.com/fsnotify/fsnotify)](https://goreportcard.com/report/github.com/fsnotify/fsnotify)
+If you are reading this, you use `master` branch of this repository,
+which is wrong.
 
-fsnotify utilizes [golang.org/x/sys](https://godoc.org/golang.org/x/sys) rather than `syscall` from the standard library. Ensure you have the latest version installed by running:
+This branch
+ - should not be used;
+ - is not maintained;
+ - is not supported;
+ - will be removed soon.
 
-```console
-go get -u golang.org/x/sys/...
+You should switch to using the default branch instead.
+
+## Using git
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,130 +1,20 @@
...
-[![GoDoc](https://godoc.org/github.com/fsnotify/fsnotify?status.svg)](https://godoc.org/github.com/fsnotify/fsnotify) [![Go Report Card](https://goreportcard.com/badge/github.com/fsnotify/fsnotify)](https://goreportcard.com/report/github.com/fsnotify/fsnotify)
+If you are reading this, you use `master` branch of this repository,
+which is wrong.
 
-fsnotify utilizes [golang.org/x/sys](https://godoc.org/golang.org/x/sys) rather than `syscall` from the standard library. Ensure you have the latest version installed by running:
+This branch
+ - should not be used;
+ - is not maintained;
+ - is not supported;
+ - will be removed soon.
```
nathany, line 10: Thanks @kolyshkin 

Do we want to say "soon"? I'm not sure when it will be removed...

nathany (commented):
Thanks for doing this. It looks reasonable, and will hopefully alleviate issues for some people until they update things.

Wondering if we should include the "how to update your local branch" text in the readme on master? Or do you think it's unnecessary (given their use case)?
https://github.com/fsnotify/fsnotify/issues/426#issuecomment-1034067268

README.md
kolyshkin: I wrote "in 2024" initially, but when I looked at it from user's perspective, I realized that "in 2024" reads like "there's nothing to worry about, you can relax for now" which is not the message I want to send.

So, I'd rather say "soon", having in mind a very broad definition of the term. :)

I can remove it if you feel it doesn't work.

nathany: That makes sense. Let's keep "soon"

kolyshkin:
> Wondering if we should include the "how to update your local branch" text in the readme on master? Or do you think it's unnecessary (given their use case)?

Makes sense, thanks, added!

PR #430, opened 2022-02-17
CI: fix golangci-lint warnings

kolyshkin:

1. Simplifies `lint` CI job by removing `go vet` and `gofmt` (those are called by golangci-lint anyway).

2. Fixes all warnings found by the default set of golangci-lint checkers. In some cases those are real bugs, so please review carefully.

3. Stop ignoring golangci-lint warnings.

Please review commit by commit, noting the commit messages.

⚠️ Currently the linters only check the code for the current platform (that is, Linux). There are multiple issues with *bsd and windows (those can be seen from CLI using e.g. `GOOS=darwin golangci-lint run`) that I am afraid to tackle.

No manual tests needed; this is why we have CI! 😉 👌🏻

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -80,7 +80,9 @@ func (w *Watcher) Close() error {
 	close(w.done)
 
 	// Wake up goroutine
-	w.poller.wake()
+	if err := w.poller.wake(); err != nil {
+		return fmt.Errorf("poller.wake failed: %w", err)
+	}
 
 	// Wait for goroutine to close
 	<-w.doneResp
--- a/inotify_poller.go
+++ b/inotify_poller.go
@@ -119,10 +119,6 @@ func (poller *fdPoller) wait() (bool, error) {
 				}
 			}
 			if event.Fd == int32(poller.pipe[0]) {
-				if event.Events&unix.EPOLLHUP != 0 {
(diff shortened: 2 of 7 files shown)


inotify.go
```diff
@@ -80,7 +80,9 @@ func (w *Watcher) Close() error {
 	close(w.done)
 
 	// Wake up goroutine
-	w.poller.wake()
+	if err := w.poller.wake(); err != nil {
+		return fmt.Errorf("poller.wake failed: %w", err)
+	}
```
arp242, line 85: I'm not entirely sure if this is really the correct behaviour; I think that this failing isn't really a big deal in the Close() here?

inotify_poller.go
```diff
@@ -119,10 +119,6 @@ func (poller *fdPoller) wait() (bool, error) {
 				}
 			}
 			if event.Fd == int32(poller.pipe[0]) {
-				if event.Events&unix.EPOLLHUP != 0 {
-					// Write pipe descriptor was closed, by us. This means we're closing down the
-					// watcher, and we should wake up.
-				}
```
arp242, line 125: The comment says it "should wake up", but it doesn't do anything. I'm not sure if this can just be removed, or if there's an omission here and we should put something in this branch.

arp242:
Should be fixed in the main branch now.

PR #444, opened 2022-04-21
go.mod: use latest x/sys

kevinburkesegment:
This version contains several upgrades including a newer version of
glibc and some bug fixes.


shogo82148 (approved):
LGTM

PR #455, opened 2022-06-03
No error when stopping watching a non watched file

MidasLamb:
Currently, when a non-watched item is requested to be removed from a
watch, an error is returned.
However, the post-condition of removing it is: "This file/dir should not
be watched anymore", which is the case also when the item wasn't being
watched in the first place.
Rather than returning an error. Return nil and skip a part of the work.

inotify.go, the Remove function definition.

We've noticed this because this lib is being used in promtail and it was logging errors because of this.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -134,7 +134,8 @@ func (w *Watcher) Remove(name string) error {
 
 	// Remove it from inotify.
 	if !ok {
-		return fmt.Errorf("can't remove non-existent inotify watch for: %s", name)
+		// Nothing to remove, as it's not present.
+		return nil
 	}
 
 	// We successfully removed the watch if InotifyRmWatch doesn't return an


mattn:
If this Remove function returns a wrapped error for `os.ErrNotExist`, will your problem be resolved with using `erros.Is`?

MidasLamb:
> If this Remove function returns a wrapped error for `os.ErrNotExist`, will your problem be resolved with using `erros.Is`?

That could also be a possibility. I'm not that experienced in go so I don't know what would be the most idiomatic way.
I can update the PR if you think it's a better approach

mattn:
I prefer to add `ErrNonExistentWatch` error in fsnotify.go

```
var (
    ErrNonExistentWatch = errors.New("can't remove non-existent watcher")
    ErrEventOverflow    = errors.New("fsnotify queue overflow")
)
```
and return it in inotify.go, kqueue.go, windows.go
```
return fmt.Errorf("%w: %s", ErrNonExistentWatch, name)
```


any thought? @nathany @shogo82148

nathany:
@mattn a new error sounds good to me

arp242:
I agree that outright removing the error is not a good solution; I opened #460 to return a specific error that can be tested for, which should address the issue better.

PR #458, opened 2022-07-03
Correct typo on readme "Usage" section

EthanEFung:

simple documentation update

README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ func main() {
 					return
 				}
 				log.Println("event:", event)
-				if event.Op&fsnotify.Write == fsnotify.Write {
+				if event.Op == fsnotify.Write {
 					log.Println("modified file:", event.Name)
 				}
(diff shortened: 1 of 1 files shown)


shogo82148:
Thank you for your pull request!

However, the current usage is correct.
Because `event.Op` may contain multiple operations.
For example, if create and write events are triggered in same time, you get `fsnotify.Create|fsnotify.Write` as `event.Op`.
In this case, `event.Op == fsnotify.Write` would miss the event.

EthanEFung:
@shogo82148 I'll take your word for it. That said, could you help me understand what `event.Op&fsnotify.Write` is doing exactly? This is the first time I've seen this sort of syntax in go and it feels a bit clever to me.

EthanEFung:
leaving this here for anyone who makes a similar pull request
https://www.reddit.com/r/golang/comments/n51nl6/could_someone_explain_what_this_is_actually_doing/

PR #460, opened 2022-07-21
Use common error when removing an unwatched file

arp242:
Applies https://github.com/shogo82148/fsnotify/pull/8 to replace #455.

---

The errors returned by the various implementations of the watcher are all different
which makes handling them difficult. This PR follows the suggestion in:
https://github.com/fsnotify/fsnotify/pull/455#issuecomment-1146037157 by @mattn
to create a common error which is wrapped by the implementations.

Replaces: https://github.com/fsnotify/fsnotify/pull/455
Signed-off-by: Andrew Thornton <art27@cantab.net>

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -65,5 +65,6 @@ func (e Event) String() string {
 
 // Common errors that can be reported by a watcher
 var (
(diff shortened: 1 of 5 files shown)


shogo82148 (approved):
Nice catch. Thanks!
LGTM

PR #462, opened 2022-07-21
Windows explicit (un)lock

jolheiser:

Fixes a locking issue in Windows

While `defer` is usually ideal for these situations, due to how some goroutines may be reading events sent to channels before the function returns I think explicit (un)locking for this specific case may be required.

Tested the GH actions over at https://github.com/jolheiser/fsnotify/pull/1


arp242 (approved):

arp242:
Cheers, thanks!

PR #465, opened 2022-07-22
inotify: fix race in Close()

### LINKED ISSUE #367: There is a race in inotify.go:watcher.Close()
Hi!

Thanks a lot for this library.

Looking at the code I'm suspicious that there is a race in Close():

https://github.com/fsnotify/fsnotify/blob/7f4cf4dd2b522a984eaca51d1ccee54101d3414a/inotify.go#L72-L80

Is it possible that two concurrent goroutines running Close() could both pass the check function `w.isClosed()` and then both call `close(w.done)`? That would result in two closes of the channel and a panic. 

One solution would be to lock the `w.mu` at least around the `w.IsClosed` and `close(w.done)` calls.

arp242:
Would sometimes fail with:

	goroutine 204 [running]:
	github.com/fsnotify/fsnotify.(*Watcher).Close(0xc0003e6410)
        	/home/martin/code/Golib/fsnotify/inotify.go:82 +0x66
	created by github.com/fsnotify/fsnotify.TestCloseRace
        	/home/martin/code/Golib/fsnotify/integration_test.go:1256 +0x1a5
	exit status 2

Because isClosed() might return "false" for two goroutines, and then
they will both try to close it, resulting in the panic.

Fixes #367

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -72,12 +72,15 @@ func (w *Watcher) isClosed() bool {
 
 // Close removes all watches and closes the events channel.
 func (w *Watcher) Close() error {
+	w.mu.Lock()
 	if w.isClosed() {
+		w.mu.Unlock()
 		return nil
 	}
 
 	// Send 'close' signal to goroutine, and set the Watcher to closed.
 	close(w.done)
+	w.mu.Unlock()
 
 	// Wake up goroutine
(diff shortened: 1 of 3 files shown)


shogo82148 (approved):
LGTM

PR #467, opened 2022-07-22
Clarify README on network drives

arp242:
It already mentioned NFS, but also mention SMB. Also rename "OS X" to "macOS" while I'm at it.


shogo82148 (approved):
LGTM

PR #463, opened 2022-07-21
Add cmd/fsnotify

arp242:
This adds a little example that can be run:

	% go run ./cmd/fsnotify **/*
	12:76:11.4710 watching; press ^C to exit
	12:76:13.6549   1 REMOVE                "cmd/fsnotify/4913"
	12:76:13.6550   2 RENAME                "cmd/fsnotify/main.go"
	12:76:13.6550   3 RENAME                "cmd/fsnotify/main.go"
	12:76:13.6612   4 CREATE                "cmd/fsnotify/main.go"
	12:76:13.6612   5 WRITE      (modified) "cmd/fsnotify/main.go"
	12:76:13.6614   6 WRITE      (modified) "cmd/fsnotify/main.go"
	12:76:13.6648   7 CHMOD                 "cmd/fsnotify/main.go"
	12:76:15.2919   8 CREATE                "cmd/fsnotify/4913"
	12:76:15.2919   9 REMOVE                "cmd/fsnotify/4913"
	12:76:15.2922  10 CHMOD                 "cmd/fsnotify/main.go"
	12:76:15.2923  11 REMOVE                "cmd/fsnotify/main.go"
	12:76:15.2925  12 RENAME                "cmd/fsnotify/main.go"
	12:76:15.2986  13 CREATE                "cmd/fsnotify/main.go"
	12:76:15.2986  14 WRITE      (modified) "cmd/fsnotify/main.go"
	12:76:15.2988  15 WRITE      (modified) "cmd/fsnotify/main.go"
	12:76:15.3024  16 CHMOD                 "cmd/fsnotify/main.go"
	^C

It's mostly the same as the example in the README, except that it takes
paths from the commandline and aligns things a bit nicer.

This is useful for documentation and experimentation. For example, while
reviewing some PRs I found it useful to quickly get an overview of "what
does fsnotify do now, and what does it do with this patch?"

PR #470, opened 2022-07-24
inotify: don't ignore events for files that don't exist

arp242:
This is quite an odd check, leading to an inconsistent event stream,
which also doesn't match what other platforms do. If you do a "CREATE +
MODIFY + REMOVE" in quick succession then you probably *want* all three
events. If you don't want to operate on non-existing files, then you can
check this in your application code.

You need to do that already, since this check is far from reliable. In
the time between this check and the application code doing something
with an event the file may have been deleted already.

I looked a bit at the history of this, and looks like it was added in
2013 with cc2c34e; issue 36 refers to this issue on the old repo, which
mentions it fixes a memory leak: https://github.com/howeyc/fsnotify/issues/36

I can't reproduce that at all; using the CLI from #463 modified to print
the memory and running:

	for i in $(seq 0 10000); { touch $i; rm $i }

Memory stays at about 100/110K in both the current main branch and this.

So I think it should be safe to remove.

### DIFF
--- a/inotify.go
+++ b/inotify.go
@@ -279,7 +279,7 @@ func (w *Watcher) readEvents() {
 			event := newEvent(name, mask)
 
 			// Send the events that are not ignored on the events channel
-			if !event.ignoreLinux(mask) {
+			if mask&unix.IN_IGNORED == 0 {
 				select {
 				case w.Events <- event:
 				case <-w.done:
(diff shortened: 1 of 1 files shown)


abursavich:
The `ignoreLinux` function showed up in a profile using 34% of the cpu time for a program I needed to be really efficient. When I came to track down what it was doing, I found this PR instead. Thanks!

PR #472, opened 2022-07-25
inotify: add recursive watcher

arp242:
This adds a recursive watcher for inotify.

Updates:  #18

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -188,9 +188,10 @@ func (e Event) String() string {
 type (
 	addOpt   func(opt *withOpts)
 	withOpts struct {
-		bufsize  int
-		op       Op
-		noFollow bool
+		bufsize    int
+		op         Op
+		noFollow   bool
+		sendCreate bool
 	}
 )
 
@@ -254,11 +255,17 @@ func withNoFollow() addOpt {
 	return func(opt *withOpts) { opt.noFollow = true }
 }
 
+// "Internal" option for recursive watches on inotify.
+func withCreate() addOpt {
+	return func(opt *withOpts) { opt.sendCreate = true }
+}
+
 var enableRecurse = false
 
 // Check if this path is recursive (ends with "/..." or "\..."), and return the
 // path with the /... stripped.
 func recursivePath(path string) (string, bool) {
+	path = filepath.Clean(path)
 	if !enableRecurse { // Only enabled in tests for now.
 		return path, false
 	}
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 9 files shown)


fsnotify.go
```diff
@@ -63,8 +73,33 @@ func (e Event) String() string {
...
 
-// Common errors that can be reported by a watcher
-var (
-	ErrNonExistentWatch = errors.New("can't remove non-existent watcher")
-	ErrEventOverflow    = errors.New("fsnotify queue overflow")
-)
+// findDirs finds all directories under path (return value *includes* path as
+// the first entry).
+func findDirs(path string) ([]string, error) {
+	dirs := make([]string, 0, 8)
```
horahoradev, line 79: What's the origin of the capacity value here?

arp242: It's basically a micro-optimisation; the performance diffence between allocting an empty slice and a slice with cap=8 is negible, but if we start from 0 then adding 8 files will require a bunch of allocations for new backing arrays + copies (it doubles every time: so 1, 2, 4, 8), which is much more expensive, comparatively speaking:

	BenchmarkAllocEmpty-4             1000000000         0.2598 ns/op      0 B/op     0 allocs/op
	BenchmarkAllocEight-4             1000000000         0.2592 ns/op      0 B/op     0 allocs/op
	BenchmarkFillEightFromEmpty-4        7502155     157.9 ns/op         112 B/op     3 allocs/op
	BenchmarkFillEightFromAlloc-4      226564515         5.270 ns/op       0 B/op     0 allocs/op

I mean, ~155ns isn't all that much, but it's just as easy to type one over the other.

Why 8 and not 4 or 16? I felt 8 was a good trade-off between potential "wasted" memory and run-time speed.

inotify.go
```diff
@@ -314,20 +352,65 @@ func (e *Event) ignoreLinux(mask uint32) bool {
 	return false
 }
 
+func (w *Watcher) isRecursive(path string) *watch {
```
horahoradev, line 355: Could we change the return type here to a bool?

inotify.go
```diff
@@ -314,20 +352,65 @@ func (e *Event) ignoreLinux(mask uint32) bool {
...
+		if mask&unix.IN_ISDIR == unix.IN_ISDIR {
+			ww := w.isRecursive(name)
+			if ww != nil {
+				//err := w.add(name, true)
+				err := w.Add(filepath.Join(name, "..."))
+				if err != nil {
+					// TODO: not sure if this has a nice error message.
+					//       Also, this path could have been removed by now;
+					//       should probably filter ENOENT or something.
+					w.Errors <- err
```
horahoradev, line 386: this could block the event loop.
I'd recommend adding a select with another case for receiving from w.done, just to make sure that we're never blocking on send to the error channel after the watcher has been closed, as in:
```
			select {
			case w.Errors <- err:
			case <-w.done:
				return
			}
```

horahoradev (commented):
LGTM aside from my comments and the remaining TODOs. If you address the feedback and TODOs, then I'll approve.

solarknight:
Is this PR still active? I really need the functionality to recursive watcher.  : )

FlorianLoch:
Hi there,
current README states there is no support for recursive watching - but seems like this MR changed that? ;)

Would be great if the README could be updated accordingly!

Thanks for adding that feature 👍

PR #474, opened 2022-07-29
debian 6 test: deal with multiple packages

arp242:
The ./cmd/fsnotify doesn't have any tests, but sooner or later we'll
have subpackages so deal with it properly now.


arp242:
Ugh, got a HTTP 429 in the CI now :-/ Guess I started too many builds. I'll try again later.

PR #477, opened 2022-07-30
Add {Event,Op}.Has()

### LINKED ISSUE #107: Documentation on how to use Op and bitwise logic
This would go on the website: https://github.com/go-fsnotify/fsnotify.org as a blog post for now, and maybe a new Documentation section in the future.

How to check Op with bitwise logic.
Multiple bits may be set.
How to wrap Op in your own type with your own helper methods.
Coalescing multiple Ops on the same Event.Name using bitwise or.

Related #76.

arp242:
Using:

	if event.Op&fsnotify.Create == fsnotify.Create {
	}

is overly verbose and awkward; a Has() method makes this much nicer:

	if event.Has(fsnotify.Create) {
	}

PR #76 was previously rejected; I think that HasCreate() etc. is
overkill, but a Has() method makes the library quite a bit more
convenient IMO.

I added it to both the Event and Op to avoid confusion; with this we can
change the Op field to be private in a future v2.

Fixes #107

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ func main() {
 					return
 				}
 				log.Println("event:", event)
-				if event.Op&fsnotify.Write == fsnotify.Write {
+				if event.Has(fsnotify.Write) {
 					log.Println("modified file:", event.Name)
 				}
 			case err, ok := <-watcher.Errors:
--- a/cmd/fsnotify/main.go
+++ b/cmd/fsnotify/main.go
@@ -42,7 +42,7 @@ func main() {
 
 				i++
 				m := ""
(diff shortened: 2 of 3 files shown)


bep (approved):
Looks good to me.

nathany (approved):

KiddoV:
What version does the method ``event.Has()`` available? It is undefined on ``v1.5.4``.

arp242:
It's not yet in any released version.

diist:
I don't get it, isn't `x&y == y` equivalent to `x == y`? https://go.dev/play/p/zIdXkFffVv3

PR #478, opened 2022-07-30
Rewrite tests

arp242:
This rewrites quite a lot of tests to be much more easily readable.

While working on #472 I wanted to check "how do renames behave now?",
and found this quite hard as most test cases were >90% "plumbing", and
seeing "what file operations does this do?" and "what events do we get?"
was not very easy.

So refactor the lot, based on some work I did in #472:

- Add a bunch of "shell-like" helper functions so you're not forever
  typing error checks, filepath.Join(), and eventSeparator(). Just
  touch(t, tmp, "file") will create a file in tmp.

- Add eventCollector type which will collect all events in in a slice,
  replacing the previous "counter". This also ensures that the Watcher
  is closed within a second (this removes a lot of duplicate code).

  This is also much more precise than merely counting events; before
  random events could get emitted but if you weren't counting those then
  you'd never know.

  Downside is that some tests are a bit more flaky now as some
  behaviours are not always consistent in various edge cases; these are
  pre-existing bugs.

- Add Events (plural) type (only for tests), and helper function to
  create this from a string like:

      REMOVE  /link
      CREATE  /link
      WRITE   /link

  Which makes seeing which events are received, diffing them, etc. much
  easier.

- Add Parallel() for most tests; reduces runtime on my system from ~12
  seconds to ~6 seconds.

All in all it reduces the integrations_test.go from 1279 lines to 405
lines and it's quite easy to see which events are expected for which
operations, which should make things a lot easier going forward.

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -1,0 +1,437 @@
+package fsnotify
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"sort"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+)
+
+type testCase struct {
+	name string
+	ops  func(*testing.T, *Watcher, string)
+	want string
+}
+
+func (tt testCase) run(t *testing.T) {
+	t.Run(tt.name, func(t *testing.T) {
+		t.Parallel()
+		tmp := t.TempDir()
+
+		w := newCollector(t)
+		w.collect(t)
+
+		tt.ops(t, w.w, tmp)
+
+		cmpEvents(t, tmp, w.stop(t), newEvents(t, tt.want))
+	})
+}
+
+// We wait a little bit after most commands; gives the system some time to sync
+// things and makes things more consistent across platforms.
+func eventSeparator() { time.Sleep(50 * time.Millisecond) }
+func waitForEvents()  { time.Sleep(500 * time.Millisecond) }
+
+// newWatcher initializes an fsnotify Watcher instance.
+func newWatcher(t *testing.T, add ...string) *Watcher {
+	t.Helper()
+	w, err := NewWatcher()
+	if err != nil {
+		t.Fatalf("newWatcher: %s", err)
+	}
+	for _, a := range add {
+		err := w.Add(a)
+		if err != nil {
+			t.Fatalf("newWatcher: add %q: %s", a, err)
+		}
+	}
+	return w
+}
+
+// addWatch adds a watch for a directory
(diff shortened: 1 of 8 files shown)


arp242:
This PR changes ... a lot. It's probably easiest to look at the new versions of integration_test.go and helpers_test.go instead of the diff:

https://github.com/fsnotify/fsnotify/blob/a9c46d1aeb606084031b22e27e03f5c5296fbbcd/integration_test.go

https://github.com/fsnotify/fsnotify/blob/a9c46d1aeb606084031b22e27e03f5c5296fbbcd/helpers_test.go

bep (dismissed):

helpers_test.go
```diff
@@ -0,0 +1,437 @@
...
+
+type Events []Event
+
+func (e Events) String() string {
+	b := new(strings.Builder)
+	for i, ee := range e {
+		if i > 0 {
+			b.WriteString("\n")
+		}
+		fmt.Fprintf(b, "%-20s %q", ee.Op.String(), strings.ReplaceAll(ee.Name, `\`, "/"))
```
mattn, line 319: filepath.ToSlash ?

arp242: Ah cheers, I didn't know about that function. Updated it.

mattn:
In some tests, it appears that goroutine is not waiting for termination. This can be caused unexpected behavior when running a series of tests, which can be difficult to figure out. I don't have strong opinion but it have better to be fixed, I think in this or another pull-request. Thanks.

arp242:
> In some tests, it appears that goroutine is not waiting for termination. This can be caused unexpected behavior when running a series of tests, which can be difficult to figure out. I don't have strong opinion but it have better to be fixed, I think in this or another pull-request. Thanks.

Which tests specifically do you mean? I think all of those that use the new `eventCollector` should be fine(?) But I didn't update all the tests top use it yet, so maybe some older ones don't.

mattn:
>Which tests specifically do you mean? I think all of those that use the new eventCollector should be fine(?) But I didn't update all the tests top use it yet, so maybe some older ones don't.

Sorry, I couldn't find the tests except two.

`func (w *eventCollector) collect(t *testing.T)`  and ` func (w *eventCollector) stop(t *testing.T) Events`

arp242:
My apologies @mattn, but I'm really confused what you mean 😅 You're saying the bug is in the eventCollector? I looked over it again, and I'm not sure if I see the problem, but I'm probably missing something.

I'll go ahead and merge this for now, so I can get ahead with some other things and add test cases and such; but let me know and I'll fix it up!

PR #479, opened 2022-07-31
kqueue: don't set up watchers on unreadable files

### LINKED ISSUE #439: Directory partially watched if a file fails to open during addWatch()
**Describe the bug**
I ran into this issue while using https://github.com/spf13/viper which uses this library as a dependency and I understand that this issue may not be related to fsnotify but rather how viper handles the error. 
Suppose there's a directory that looks like this:
```
dir/
|- a.txt <- watched
|- b/ <- unix.Open returns EPERM 1 Operation not permitted
|- c.yaml <- file I'm interested in, not watched
```
And these 2 things happen:
1. c.yaml is never reached
2. The only event that's ever going to be fired is "b/ is created"

So in viper a watch for a directory is added here https://github.com/spf13/viper/blob/v1.10.1/viper.go#L450. During `Add()`, the code loops through all the files under dir/ here https://github.com/fsnotify/fsnotify/blob/v1.5.1/kqueue.go#L398, which include a.txt, b/, and c.yaml, in order. 
```
	for _, fileInfo := range files {
		filePath := filepath.Join(dirPath, fileInfo.Name())
		filePath, err = w.internalWatch(filePath, fileInfo)
		if err != nil {
			return err
		}

		w.mu.Lock()
		w.fileExists[filePath] = true
		w.mu.Unlock()
	}
```
When `w.internalWatch()` is called for b/, an err is returned, which causes Add() to return with err. However, the watcher is already updated with `w.fileExists["dir/a.txt"] = true`. On the other end of the watcher, the `w.readEvents()` goroutine loops over all the files under dir/ again to discover new files. Relevant code here: https://github.com/fsnotify/fsnotify/blob/v1.5.1/kqueue.go#L444
```
func (w *Watcher) sendFileCreatedEventIfNew(filePath string, fileInfo os.FileInfo) (err error) {
	w.mu.Lock()
	_, doesExist := w.fileExists[filePath]
	w.mu.Unlock()
	if !doesExist {
		// Send create event
		select {
		case w.Events <- newCreateEvent(filePath):
		case <-w.done:
			return
		}
	}

	// like watchDirectoryFiles (but without doing another ReadDir)
	filePath, err = w.internalWatch(filePath, fileInfo)
	if err != nil {
		return err
	}
```
It first sees a.txt, which `doesExist`. It then sees b/. b/'s `doesExist` is false so a new event is fired. When the code runs to `filePath, err = w.internalWatch(filePath, fileInfo)`, it errors again and returns early without going through the rest of the directory.

**To Reproduce**

**Expected behavior**
I think ideally for my use case, either make `watchDirectoryFiles()` and `sendDirectoryChangeEvents()` continue if a file fails to open, or have another map to track files that fail to open so it doesn't keep firing off events about them.

macOS: sw_vers
```
ProductName:	macOS
ProductVersion:	11.6.4
BuildVersion:	20G417
```

**Additional context**
If applicable, add screenshots or a code sample to help explain your problem.

arp242:
Watching an entire directory would fail if it contains a file we can't
read due to permission errors. That's obviously not ideal, so just skip
the file.

Also add a test for trying to watch a directory where the permissions
are denied, and test the error is what's expected; this prevents
regressions like in #393.

Fixes #439

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -42,7 +42,8 @@ jobs:
           usesh: true
           prepare: pkg install -y go
           run: |
-            go test -race ./...
+            pw user add -n action -m
+            su action -c 'go test -race ./...'
 
   testOpenBSD:
     runs-on: macos-12
@@ -54,18 +55,18 @@ jobs:
         uses: vmactions/openbsd-vm@v0.0.6
         with:
           prepare: pkg_add go
+          # No -race as the VM doesn't include the comp set.
+          #
+          # TODO: should probably add this, but on my local machine the tests
+          #       time out with -race as the waits aren't long enough (OpenBSD
+          #       is kind of slow), so should probably look into that first.
+          #       Go 1.19 is supposed to have a much faster race detector, so
+          #       maybe waiting until we have that is enough.
           run: |
             # Default of 512 leads to "too many open files".
             ulimit -n 1024
-
-            # No -race as the VM doesn't include the comp set.
-            #
-            # TODO: should probably add this, but on my local machine the tests
-            #       time out with -race as the waits aren't long enough (OpenBSD
-            #       is kind of slow), so should probably look into that first.
-            #       Go 1.19 is supposed to have a much faster race detector, so
-            #       maybe waiting until we have that is enough.
-            go test ./...
+            useradd -mG wheel action
+            su action -c 'go test ./...'
 
   testNetBSD:
     runs-on: macos-12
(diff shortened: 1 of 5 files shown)


arp242:
Not sure why the BSD tests are failing; for example for FreeBSD:

	--- FAIL: TestWatch (0.00s)
		--- FAIL: TestWatch/file_in_directory_is_not_readable (0.98s)
			helpers_test.go:33: 
				have:
					WRITE                "/file"
					REMOVE               "/file"
					REMOVE               "/file-unreadable"
				want:
					WRITE                "/file"
					REMOVE               "/file"
	--- FAIL: TestAdd (0.00s)
		--- FAIL: TestAdd/permission_denied (0.22s)
			integration_test.go:504: error is nil

But if I test locally it's fine 🤔

	[~cg/fsnotify](kqueue-unreadable)% go-on freebsd go test -race
	go-on: freebsd already started; nothing to do
	FreeBSD 13.1-RELEASE
	PASS
	ok      github.com/fsnotify/fsnotify    9.030s

I think the chmod is not running properly for some reason.

arp242:
Oh, the NetBSD failure has a clue:

    Pseudo-terminal will not be allocated because stdin is not a terminal.
    We recommend that you create a non-root account and use su(1) for root access.

So I guess the tests are always run as root, and root can always access everything :-/

I fixed it for FreeBSD; need to look at the others.

PR #334, opened 2020-04-20
kqueue: synchronously close fd in Close()

dottedmag:
This makes it possible to create and destroy watchers quickly on macOS (e.g. in
unit tests).

Closes #333

PR #480, opened 2022-08-01
kqueue: remove timeout from unix.Kevent()

### LINKED ISSUE #89: watcher.Close() is hanging forever on OSX

### LINKED ISSUE #237: High CPU usage

### LINKED ISSUE #333: macOS: Close() does not release kqueue fd synchronously

macOS: 10.14.6

The following test fails:
```
func TestOpenCloseQuickly(t *testing.T) {
	name := tempMkFile(t, "")

	for i := 0; i < 1000; i++ {
		w := newWatcher(t)
		err := w.Add(name)
		if err != nil {
			t.Fatal(err)
		}
		err = w.Close()
		if err != nil {
			t.Fatal(err)
		}
	}
}
```
with
```
=== RUN   TestOpenCloseQuickly
    TestOpenCloseQuickly: integration_test.go:51: failed to create test file: open /var/folders/bw/6hyp_vyj68v973qbhlbcvlcr0000gn/T/fsnotify232022216: too many open files
--- FAIL: TestOpenCloseQuickly (0.00s)
```

The problem is that `w.Close()` returns before reader goroutine closes the kqueue fd.

arp242:
The timeout for unix.Kevent() is causing issues; every 100ms it will do a new
unix.Kevent() syscall, which isn't too efficient: even if you have just one
change an hour, you will still keep calling kevent() ten times per second,
resulting in a needlessly high CPU usage.

Without a timeout, kevent() will block indefinitely until there are some events,
which is much more efficient. We can't just remove the timout however, since we
can't interrupt the kevent() call on FreeBSD and NetBSD, and it will hang
forever. This issue is described in more detail here: 
https://github.com/fsnotify/fsnotify/pull/262#issuecomment-1201655890

To solve this we register a new kevent() with the file descriptor set to the
closepipe; when this pipe is closed an event is sent and kevent() will stop
blocking.

This is a rebased version of #124.

Fixes #89
Fixes #237
Fixes #333
Supersedes and closes #124
Supersedes and closes #262
Supersedes and closes #334


dottedmag:
Please add `TestOpenCloseQuickly` from #334. I have checked that on macOS 12.4 it still fails on `master`, and passes on this branch.

Updated test:
```
// closing a watcher should synchronously close watch file descriptor
func TestOpenCloseQuickly(t *testing.T) {
	tmp := t.TempDir()
	name := filepath.Join(tmp, "testfile")

	touch(t, name)

	for i := 0; i < 10000; i++ {
		// if file descriptors are not closed synchronously
		// this will fail with "too many open files"
		w := newWatcher(t)
		err := w.Add(name)
		if err != nil {
			t.Fatal(err)
		}
		err = w.Close()
		if err != nil {
			t.Fatal(err)
		}
	}
}
```

arp242:
> Please add TestOpenCloseQuickly from #334.

I added this in a local `test-resource-usage` branch that I've been working on and also tests some other things, but I haven't finished/pushed that yet.

The problem is that it takes a very long time to run on various platforms: ~25 seconds on Linux, ~40 seconds on OpenBSD (because OpenBSD is slow), etc. and that it doesn't *really* test anything except by "brute force". I'm pretty sure we can write a better test for this, but need to look at this and I don't think it's needed *right now*.

dottedmag:
I'm fine with it being Darwin-specific for now. On macOS it passes in ~500ms.

fjl:
6 years in the making... I'm interested to see if this PR will finally be merged :)

arp242:
> 6 years in the making... I'm interested to see if this PR will finally be merged :)

Yes, it will. The only reason it's not yet is because I wanted to give other people (e.g. fsnotify members) a chance to review it, but I'll merge it after the weekend regardless if no one does.

Going forward, PRs should be reviewed and merged (or rejected if it's no good) in a reasonable timeframe (days or weeks, rather than years).

PR #482, opened 2022-08-02
all: always use a pointer to sync.Mutex

### LINKED ISSUE #164: fatal error: concurrent map read and map write

arp242:
You can't copy a mutex (well, you can, but then you create a new mutex
which loses the point of having one), so use a pointer, just in case
people pass the Watcher by value.

Note that "go vet" will warn about this:

	% go vet
	# github.com/fsnotify/fsnotify
	./integration_test.go:532:9: variable declaration copies lock value to ww: github.com/fsnotify/fsnotify.Watcher contains sync.Mutex

But there's no real downside to using a pointer here.

Fixes #164

### DIFF
--- a/helpers_test.go
+++ b/helpers_test.go
@@ -255,12 +255,12 @@ func chmod(t *testing.T, mode fs.FileMode, path ...string) {
 type eventCollector struct {
 	w      *Watcher
 	events Events
-	mu     sync.Mutex
+	mu     *sync.Mutex
 	done   chan struct{}
 }
 
 func newCollector(t *testing.T) *eventCollector {
-	return &eventCollector{w: newWatcher(t), done: make(chan struct{})}
+	return &eventCollector{w: newWatcher(t), mu: new(sync.Mutex), done: make(chan struct{})}
 }
 
 func (w *eventCollector) stop(t *testing.T) Events {
--- a/inotify.go
+++ b/inotify.go
@@ -27,7 +27,7 @@ type Watcher struct {
 	fd          int
 	Events      chan Event
 	Errors      chan error
-	mu          sync.Mutex // Map access
+	mu          *sync.Mutex // Map access
 	inotifyFile *os.File
 	watches     map[string]*watch // Map of inotify watches (key: path)
 	paths       map[int]string    // Map of watched paths (key: watch descriptor)
@@ -47,6 +47,7 @@ func NewWatcher() (*Watcher, error) {
 
 	w := &Watcher{
 		fd:          fd,
+		mu:          new(sync.Mutex),
 		inotifyFile: os.NewFile(uintptr(fd), ""),
 		watches:     make(map[string]*watch),
 		paths:       make(map[int]string),
--- a/integration_test.go
+++ b/integration_test.go
@@ -523,3 +523,15 @@ func TestRemove(t *testing.T) {
 		}
 	})
 }
+
+// Make sure we don't race if people pass the watcher by value.
+func TestWatcherByValue(t *testing.T) {
+	var (
+		tmp = t.TempDir()
+		w   = newWatcher(t)
+		ww  = *w
+	)
+	for i := 0; i < 10; i++ {
+		go ww.Add(tmp)
+	}
+}
--- a/kqueue.go
+++ b/kqueue.go
@@ -27,7 +27,7 @@ type Watcher struct {
 
 	kq int // File descriptor (as returned by the kqueue() syscall).
 
-	mu              sync.Mutex                  // Protects access to watcher data
+	mu              *sync.Mutex                 // Protects access to watcher data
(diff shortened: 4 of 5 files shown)


shogo82148 (changes requested):
Thank you, but I can't merge it.
We should use `*Watcher` instead.

arp242:
> We should use *Watcher instead.

We already do, but people can dereference the pointer:

   // ...
   someFunc(*w)

This is arguably a silly thing to do, but not impossible.

This guarantees it's always correct, with basically no downsides as far as I can see other than using an extra int of memory for the pointer, which is negligible.

It came up over here, although I don't think it's the source of the reported issue: https://github.com/fsnotify/fsnotify/issues/164#issuecomment-247165616

shogo82148:
> This guarantees it's always correct, with basically no downsides as far as I can see other than using an extra int of memory for the pointer, which is negligible.

No, applying this pull request does not guarantee that it will always be correct.
Because `Watcher` might have values that are not reference type.

https://github.com/fsnotify/fsnotify/blob/a566bb1ede15a7a35b8563824e361d8e1a57f353/kqueue.go#L37

The only correct way is to always use `*Watcher`.

arp242:
But `isClosed` is protected by the mutex, so copying that is fine, right? Or am I completely misunderstanding some things here?

arp242:
The line:

	github.com/fsnotify/fsnotify/inotify.go:253

Corresponds to this in `readEvents()` in v1.3.1:

	w.mu.Lock()
	name := w.paths[int(raw.Wd)]
	w.mu.Unlock()

So it *is* locked, but it looks like you're derefencing the pointer per oozie's example?

Basically, don't do that 🙃 But I created #482 to fix it.

arp242:
> It came up over here, although I don't think it's the source of the reported issue: #164 

Actually, it does seem to be the source of #164 after investigating more.

Reopening so we can discus more, because while I think dereferencing the pointer is a silly and perhaps even a stupid thing to do, I think we should make the API as "safe" as possible, and unless I misunderstood things this should allow for safe dereferencing.

shogo82148:
> But `isClosed` is protected by the mutex, so copying that is fine, right?

No, it's not fine.
The original `isClosed` is protected by the mutex, however copied one is not.

For example, here is the code simplified for illustration:

https://go.dev/play/p/1sEro3PrnJj

```go
package main

import (
	"sync"
)

type Watcher struct {
	mu       *sync.Mutex
	isClosed bool
	done     chan struct{}

	// ... (snip) ...
}

func NewWatcher() *Watcher {
	return &Watcher{
		mu:   new(sync.Mutex),
		done: make(chan struct{}),
	}
}

func (w *Watcher) Close() error {
	w.mu.Lock()
	defer w.mu.Unlock()
	if w.isClosed {
		return nil
	}
	w.isClosed = true

	// ... (snip) ...

	close(w.done)
	return nil
}

func main() {
	w1 := NewWatcher()
	w2 := *w1
	w1.Close()
	w2.Close()
}
```

it panics.

```

goroutine 1 [running]:
main.(*Watcher).Close(0x404bf8?)
	/tmp/sandbox2491312479/prog.go:32 +0xbe
main.main()
	/tmp/sandbox2491312479/prog.go:40 +0x8f

Program exited.
```

arp242:
Okay, it seems I don't fully understand how this works, or rather, it works different than I had always assumed; need to look a bit deeper into this to get a more accurate understanding. Thank you for explaining.

PR #483, opened 2022-08-03
ci: use QEMU to test different architectures

arp242:


arp242:
On second thought, I don't really know how accurat QEMU user-mode is in emulating "real-world" conditions; all the system calls and asm opcodes are translated in real-time, which is useful and enough to run applications, but not really identical to actually running the application on e.g. Linux/arm.

e.g. the 386 test fails with *Weird Shit™* when `GOARCH=386 go test -c; ./fsnotify.test` works fine; it seems like some of the syscalls/assembly gets translated by QEMU in an unexpected way.

So yeah, never mind for now. Should use `qemu-system-*`, but that's a lot slower so meh.

Issue #72, opened 2015-03-09
short read in readEvents on Windows


bbigras:
I have some `watcher.Errors: short read in readEvents()` once every couple of minutes.

I'm not able to reproduce the problem if I run the smallest code possible to watch the same directory.

What should I try next to reproduce the `short read` error?

go version go1.4.2 windows/386

nathany:
That error would be coming from here: https://github.com/go-fsnotify/fsnotify/blob/master/windows.go#L448

Unfortunately I don't know the Windows code very well yet. It was inherited from Chris Howey's version of fsnotify, which was inherited from [winfsnotify](https://github.com/golang/exp/tree/master/winfsnotify).

If you manage to add enough to a code example to reproduce the issue, that would really help.

rubenhazelaar:
I have a similar problem (Windows 8.1), running golang 1.5.2 and the latest go-fsnotify (from gopkg.in/fsnotify.v1). When I place a lot of files (200+) in my directory I get the following:

```
2016/02/10 15:08:31 error: short read in readEvents()
2016/02/10 15:08:35 error: short read in readEvents()
2016/02/10 15:08:37 error: short read in readEvents()
2016/02/10 15:08:40 error: short read in readEvents()
2016/02/10 15:08:43 error: short read in readEvents()
2016/02/10 15:08:46 error: short read in readEvents()
2016/02/10 15:08:49 error: short read in readEvents()
2016/02/10 15:08:51 error: short read in readEvents()
2016/02/10 15:08:58 EOF
```

When EOF is put out, the executable crashes. Could this be something like a buffer overflow?

nathany:
Sorry, I'm still not really sure. :disappointed:

rubenhazelaar:
I did some testing. The EOF was caused by a xml decoder I use, so solved that one.

However the short read error remains. 

In my setup I use ftp to upload a bunch of files. The problem also occurs when I just copy the files in the folder. When a create event is fired (it fires twice btw, don't know if this is normal) I decode the content of each file and put that content in a db. So quite heavy work. Could this be interfering with fsnotify's ability to process all the new files? Or perhaps because it's a lot to process the os takes a different route which throws off fsnotify or the underlying API? Just speculating btw :)

rubenhazelaar:
Ok, had the opportunity to test on a Linux machine and had a weird difference:

On Windows the WRITE event is fired **3x**, in Linux (ElementaryOS) **1x**

[...]

PR #485, opened 2022-08-03
windows: increase buffer size from 4K to 64K

arp242:
People are running in to trouble because the 4K buffer can overflow;
see: #72.

This is not a "real" fix, but I think a 64K buffer is acceptable even on
memory-limited machines; no one is running the Windows on an Arduino,
and even with something like 128M of memory, the extra 124K is basically
negligible. There is also no real performance difference between
allocating a large array vs. a small array: they're both comparable.

It should probably be enough for most applications. Need to look in the
future to either dynamically grow the size, or allow setting it similar
to what tilt does as mentioned in #72.

### DIFF
--- a/windows.go
+++ b/windows.go
@@ -195,7 +195,7 @@ type watch struct {
 	mask   uint64            // Directory itself is being watched with these notify flags
 	names  map[string]uint64 // Map of names being watched and their notify flags
 	rename string            // Remembers the old name while renaming a file
-	buf    [4096]byte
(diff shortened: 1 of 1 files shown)


nshalman (approved):
Seems reasonable to me, with the caveat that I don't have deep insight into Windows.

arp242:
> I don't have deep insight into Windows.

Yeah, me neither, but tilt has been running a fork of this with a 64k buffer size for over two years, so it's actually fairly well tested.

Issue #364, opened 2021-03-13
possible key buffer overflow when calling fsnotify/fsnotify on Windows


AZ-X:
Before reporting an issue, please ensure you are using the latest release of fsnotify.

### Which operating system (GOOS) and version are you using?

Windows:  ALL (7, 8, 8.1 10)

### Please describe the issue that occurred.

golang/go#44593
See https://github.com/fsnotify/fsnotify/blob/7f4cf4dd2b522a984eaca51d1ccee54101d3414a/windows.go#L381

PR #486, opened 2022-08-04
windows: replace syscall with golang.org/x/sys/windows

arp242:
Specifically to address https://github.com/fsnotify/fsnotify/issues/364; syscall.GetQueuedCompletionStatus()
contains a bug that can't be fixed in a backwards-compatible way, so
they didn't. Once you use sys/windows for one thing, you gotta use it
for everything.

I ran the tests with -count=100 locally, and it all seems fine.

While I'm here, also replace "e" with "err" for error variables.
The way it was used now was sometimes also confusing.

Fixes https://github.com/fsnotify/fsnotify/issues/364

### DIFF
--- a/windows.go
+++ b/windows.go
@@ -16,8 +16,9 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"syscall"
 	"unsafe"
+
+	"golang.org/x/sys/windows"
 )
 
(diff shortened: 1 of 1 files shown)


mattn (approved):
LGTM

PR #495, opened 2022-08-08
Correct readme example

dpaks:
I have signed the Google CLA in 2020.

The example given in Readme is wrong as no one consumes the unbuffered channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ func main() {
 
 	done := make(chan bool)
 	go func() {
+	        defer close(done)
 		for {
 			select {
 			case event, ok := <-watcher.Events:


arp242:
It's just there to block the main goroutine so it doesn't immediately exit; that close() will never get called. I'll add a comment to clarify (was working some docs issues yesterday anyway).

shogo82148:
```go
<-done
```

it can be replace with:

```go
select {} // block forever
```

https://go.dev/ref/spec#Select_statements

arp242:
I replaced it with this in the branch I'm working on:

    // Block main goroutine forever.
    <-make(chan struct{})

Which is also what I used in the examples I added.

No need to make a `done` variable, but other than that: I don't think it really matter what you use? I've mostly seen a channel used for this over the years.

arp242:
I merged my branch; I don't think there's a good reason to prefer one method over the other? So I'll close this for now.

PR #497, opened 2022-08-10
update the basic example on readme

isaaguilar:
`fsnotify.Event` does not have the `Has` method. I'm not sure what `Has` was supposed to do, but I'm guessing that it might be equivalent to using the `==` for the Op of event

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ func main() {
 					return
 				}
 				log.Println("event:", event)
-				if event.Has(fsnotify.Write) {
+				if event.Op == fsnotify.Write {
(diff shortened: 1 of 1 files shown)


arp242:
It was added two weeks ago, but it's not in any release (yet): https://github.com/fsnotify/fsnotify/blob/main/fsnotify.go#L74-L75

You probably want to use https://pkg.go.dev/github.com/fsnotify/fsnotify, which also includes the README of that version with the (old) example.

arp242:
Added a note to the README that it reflects the latest dev version, pointing people to pkg.go.dev for the latest released version of it.

PR #506, opened 2022-09-12
Create dependabot.yml

shogo82148:


arp242:
We have one dependency, and one that doesn't even get any tagged releases. I'm not sure what this adds?

shogo82148:
`package-ecosystem: "gomod"` is not necessary, however we need `package-ecosystem: "github-actions"` section in `dependabot.yml`.
We have some dependencies in GitHub Actions workflows.
For example, `.github/workflows/test.yml` depends to the [freebsd-vm](https://github.com/marketplace/actions/freebsd-vm) action:

https://github.com/fsnotify/fsnotify/blob/85acde25252d3a042b11a58e13e53b2b9ac343de/.github/workflows/test.yml#L68

and here is newer version https://github.com/vmactions/freebsd-vm/releases/tag/v0.2.6

arp242:
If it works with the current version then it's okay, no?

I'm not really looking forward to testing updates for stuff like this unless it's broken.

PR #510, opened 2022-09-20
readme: fix example

frzifus:
When running the example, i got the following error message:
```
event.Has undefined (type fsnotify.Event has no field or method Has)
```


arp242:
> It's best to read the documentation at pkg.go.dev, as it's pinned to the last released version, whereas this README is for the last development version which may include additions/changes.

PR #512, opened 2022-09-28
fixed README.md example to use the Op property instead of the Has method

Davidonium:


arp242:
From the README:

> It's best to read the documentation at pkg.go.dev, as it's pinned to the last released version, whereas this README is for the last development version which may include additions/changes.

Davidonium:
Apologies for the noise, looks like I didn't properly read what I was trying to fix 😅

PR #516, opened 2022-10-13
Return ErrClosed on Add() when the watcher is closed

### LINKED ISSUE #511: Make watcher `isClosed` public
**Is your feature request related to a problem? Please describe.**
In some cases I am trying to add files to the watcher while the watcher has already been closed. Instead of writing logic on the (OS dependent) error message I receive, I want to check whether the watcher has been closed myself before adding files.

**Describe the solution you'd like**
Make `Watcher.isClosed` public.

**Describe alternatives you've considered**
```
...
err := watcher.Add(path)
if err.Error() == "inotify instance already closed" || err.Error() == "watcher already closed" { // OS dependent
    ...
}
```

**Additional context**

arp242:
Also consistently return nil for Remove() and WatchList(). For Remove() we could return an error, but I don't really see the point: if the watcher is closed it's basically a no-op so returning nil is fine.

For WatchList() I'd like to return an error, but there is no error return, so explicitly returning nil is the best we can do.

Fixes #511

### DIFF
--- a/backend_windows.go
+++ b/backend_windows.go
@@ -120,9 +120,9 @@ type Watcher struct {
 	input chan *input    // Inputs to the reader are sent on this channel
 	quit  chan chan<- error
 
-	mu       sync.Mutex // Protects access to watches, isClosed
-	watches  watchMap   // Map of watches (key: i-number)
-	isClosed bool       // Set to true when Close() is first called
+	mu      sync.Mutex // Protects access to watches, closed
+	watches watchMap   // Map of watches (key: i-number)
+	closed  bool       // Set to true when Close() is first called
 }
 
 // NewWatcher creates a new Watcher.
@@ -141,6 +141,12 @@ func NewWatcher() (*Watcher, error) {
 	}
 	go w.readEvents()
 	return w, nil
+}
+
+func (w *Watcher) isClosed() bool {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	return w.closed
 }
 
 func (w *Watcher) sendEvent(name string, mask uint64) bool {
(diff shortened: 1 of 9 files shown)


arp242:
CI failures are unrelated to this PR.

fsnotify_test.go
```diff
@@ -892,6 +892,28 @@ func TestClose(t *testing.T) {
...
+		file := filepath.Join(tmp, "file")
+		touch(t, file)
+		if err := w.Add(file); !errors.Is(err, ErrClosed) {
+			t.Fatalf("wrong error for Add: %#v", err)
+		}
+		if err := w.Remove(file); err != nil {
+			t.Fatalf("wrong error for Remove: %#v", err)
+		}
+		if ugh := w.WatchList(); ugh != nil { // This function is crap.
+			t.Fatalf("WatchList not nil: %#v", ugh)
```
nshalman, line 914: Probably better to use err instead of ugh. That said, I know the feeling!

backend_windows.go
```diff
@@ -244,6 +246,13 @@ func (w *Watcher) Add(name string) error {
 //
 // Removing a path that has not yet been added returns [ErrNonExistentWatch].
 func (w *Watcher) Remove(name string) error {
+	w.mu.Lock()
+	if w.isClosed {
```
nshalman, line 250: Should the Windows isClosed be modified to lock the Mutex, or are there places where that could cause deadlock if we did?

nshalman (approved):
Some nits, but this seems reasonable to me.
Thanks for fixing the illumos CI!

backend_windows.go
nshalman: Fixed in f8aebb4. Nice!

fsnotify_test.go
nshalman: Ah, I get it now. Thanks for the improvement in f8aebb4!

nshalman (approved):

fsnotify_test.go
arp242: I just get annoyed every time I see this function because everything about it is wrong IMO, which is impressive for such a small feature 🙃

PR #518, opened 2022-10-13
inotify: remove watch when renaming a watched path

### LINKED ISSUE #172: Problem when moving watched directory with inotify
When renaming/moving a watched directory that is not inside a watched directory, fsnotify informs of the renaming, but:
1. doesn't update the name stored internally
2. keeps tracking changes inside the directory
3. report names starting with the old name for changes inside it 
4. attempt to remove the watch will fail wit an error

This is deduced from reading the code and testings with inotify. I didn't verify if this happens and how fsnotify would precisely behave (e.g 4).  I'm confident that this is the case.

The reason is due to inotify not returning the new name of the watched directory (IN_MOVE_SELF). If the watched directory is inside a watched directory than the new name may be obtained, but this needs some smart operation. If that watched directory is moved out of the watched directory, it is still watched, bu the new name is unknown and the watcher can't be removed. 

One possible fix is to automatically unwatch moved/renamed directories that aren't inside a watched directory. Watched directories inside another watched directories should be unwatched only when moved out of the enclosing watched directory. This is required because it is not possible to know their new name. 

It is yet unclear if it is possible to track directory moves from one watched directory to another watched directory (could be a subdirectory ) I need to do further tests.

### LINKED ISSUE #503: Rename, and re-Add doesn't work
```
watcher.Add("foo")
// mv foo bar  //  (RENAME event is generated)
watcher.Add("foo")
// write to "foo"
// no events for "foo"
```

Linux: 6.0.0-rc4

arp242:
Renaming a watched path is problematic; on inotify we just get a IN_MOVED_SELF event with the old filename and that's it; no more events for you! So if you do:

	watch one
	mv    one two
	echo  asd >two

You still continue to get events for the file "one", even though it's now named "two" (the file descriptor doesn't care about the rename). There is no way we can know the new event as far as I can tell, inotifywait(1) also behaves like this. So instead of continuing in a semi-broken state just remove the watcher, like we do for deletes.

On kqueue and FEN the situation is similar, and we actually already removed watchers on renames.

On Windows this all works nicely; the watch is preserved and the filename is updated. I decided to keep this as-is for now, even though it's inconsistent. We actually fixed the Windows behaviour for the 1.6.0 release in #370 , so people do seem to care about it and use it, and experience with the symlink change in 1.5.0 shows it's better to keep inconsistent behaviour rather than change it. Fixing this up is something for a v2.

Fixes #172
Fixes #503

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,10 +13,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changes and fixes
 
+- inotify: remove watcher if a watched path is renamed ([#518])
+
+  After a rename the reported name wasn't updated, or even an empty string.
+  Inotify doesn't provide any good facilities to update it, so just remove the
+  watcher. This is already how it worked on kqueue and FEN.
+
+  On Windows this does work, and remains working.
+
 - all: return ErrClosed on Add() when the watcher is closed ([#516])
 
 [#371]: https://github.com/fsnotify/fsnotify/pull/371
 [#516]: https://github.com/fsnotify/fsnotify/pull/516
+[#518]: https://github.com/fsnotify/fsnotify/pull/518
 
 ## [1.6.0] - 2022-10-13
 
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -187,11 +187,11 @@ func (w *Watcher) Close() error {
 //
 // A path can only be watched once; attempting to watch it more than once will
 // return an error. Paths that do not yet exist on the filesystem cannot be
-// added. A watch will be automatically removed if the path is deleted.
+// added.
 //
-// A path will remain watched if it gets renamed to somewhere else on the same
-// filesystem, but the monitor will get removed if the path gets deleted and
-// re-created, or if it's moved to a different filesystem.
+// A watch will be automatically removed if the watched path is deleted or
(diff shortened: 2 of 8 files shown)


nshalman (approved):
One nit on a comment.

Between the tests passing and my review, I think we're good.
I do think I have a pretty good handle on the changes you made to the tests, but all the same I'd prefer if someone else with more inotify experience approved too.

That said, if folks are busy, I wouldn't wait an inordinate amount of time for a second review.

arp242:
> I'd prefer if someone else with more inotify experience approved too. That said, if folks are busy, I wouldn't wait an inordinate amount of time for a second review.

You're basically the only person regularly doing reviews at the moment, so ... 🙃 Even at $dayjob getting people to review things in a vaguely timely manner has often been a challenge.

My general approach has been "just merge it if I'm fairly confident it's good". People can always comment later, or we can fix things later, or even backout something if need be.

I tend to work on these kind of things in "bursts", e.g. a lot for a week and then nothing for a bit while I do other stuff. I don't really expect people to review >10 PRs in a week for a project like this.

I think I will just stop asking them except for the larger changes like #521 that change the API; technical fixes like this one are easy to prove correct, but API changes and the like are a more subjective matter of taste, and it's good to have some people look at that much more than stuff like this. That would reduce the requests, improving the chance people actually look at it.

PR #520, opened 2022-10-13
windows: ignore file attributes

### LINKED ISSUE #487: Consider ignoring CHMOD on Windows
Tilt includes the following change in their fork: https://github.com/fsnotify/fsnotify/compare/main...tilt-dev:fsnotify:main

> Windows allows filtering file events to include changes to file
> attributes. However, when reporting the actual event, attribute
> changes are included as `FILE_ACTION_MODIFIED` so it's impossible
> to distinguish from file writes.
> 
> This means the library's `op::Chmod` is never actually populated
> on Windows and so attribute changes aren't really usable. Since
> the API does not support specifying an event type filter as a
> caller, attribute changes on Windows are ignored rather than being
> misreported.

and the README says:

> * [Windows] File attribute changes are ignored
>   * No `Chmod` operations will be detected or returned on Windows because the underlying Windows API does not support
>     distinguishing these from `Write` operations
>   * Some software (likely AV/security) can cause excessive attribute changes resulting in many spurious write events
>     for otherwise unchanged files

Might want to consider including that here; it seems these events are never useful, and only cause problems. Need to look into it.

arp242:
On Windows a FILE_ACTION_MODIFIED event (i.e. a Write event) is triggered on file attribute changes, rather than some dedicates ATTRIBUTE_CHANGED event. Looking at the docs, I don't really see a way to distinguish between "real" write events and attribute changes. This is very odd, but seems to be how the ReadDirectoryChangesW() API works.

The only way I can see to distinguish between the two events is to set up two filters: one with a FILE_NOTIFY_CHANGE_ATTRIBUTES, and one without. But that seems overly complex, and no one asked to get Chmod events for Windows; it's not really all that interesting on Windows anyway.

The problem is that some software (anti-virus, backup software, etc.) can issue lots of attribute changes, causing a lot of "fake" Write events.

So remove the FILE_NOTIFY_CHANGE_ATTRIBUTES and sysFSATTRIB flags.

This was adapted from the tilt-dev/fsnotify fork:
https://github.com/tilt-dev/fsnotify/pull/8

Fixes #487

### DIFF
--- a/backend_windows.go
+++ b/backend_windows.go
@@ -294,7 +294,6 @@ func (w *Watcher) WatchList() []string {
 // This should all be removed at some point, and just use windows.FILE_NOTIFY_*
 const (
 	sysFSALLEVENTS  = 0xfff
-	sysFSATTRIB     = 0x4
 	sysFSCREATE     = 0x100
 	sysFSDELETE     = 0x200
 	sysFSDELETESELF = 0x400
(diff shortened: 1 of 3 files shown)


arp242:
FYI @milas @nicks

PR #523, opened 2022-10-14
Add WithEvents to filter events

### LINKED ISSUE #7: Built-in event filtering of Ops
The current mechanism for filtering events is to inspect the Op bitmask:

``` go
if event.Op&fsnotify.Create == fsnotify.Create {
```

In [howeyc/fsnotify](https://github.com/howeyc/fsnotify) there is a `WatchFlags` method to filter operations. WatchFlags was removed in 10e144023ee0d6da2b589d0cb767c06d0c3d5cdb (v0.10.0) because:
- The implementation required a lot of extra bookkeeping while providing little benefit over filtering events as they were received.
- As @mstum reported, it wasn't working on Windows, which was due to the bookkeeping not being implemented for files within a directory watch: https://github.com/howeyc/fsnotify/issues/93#issuecomment-39285195.
- There were no tests for WatchFlags.

@nightlyone [pointed out](https://groups.google.com/d/msg/golang-dev/bShm2sqbrTY/O8kaRj1zi88J) that it still is desirable to specify filters so the kernel doesn't need to wakeup our thread/process for an event we'll just ignore.

Some research ([API doc](http://goo.gl/MrYxyA)) reveal differences from one OS to the next:
- kqueue doesn't have a Create Op at all (`NOTE_CREATE`), much less a filter for it.
- Windows has a filter for `FILE_NOTIFY_CHANGE_FILE_NAME` that covers Create, Remove and Rename. An event does indicate which one of these it is, but some user-space filtering is still necessary.

arp242:
Only tested on Linux; support for other systems is hypothetical. Very much WIP.

Based on #521 rather than main.

Fixes #7

### DIFF
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -258,7 +258,7 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 	// Associate all files in the directory.
 	if stat.IsDir() {
-		err := w.handleDirectory(name, stat, true, w.associateFile, with.events)
+		err := w.handleDirectory(name, stat, true, with.events, w.associateFile)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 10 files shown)


arp242:
Not sure if we want to go this route; not adding the flags is somewhat complex, and not sure if it really gives us an advantage in terms of performance. Might as well just filter them when we send them on the Event channel.

PR #528, opened 2022-10-16
Add CI for linux/arm64, darwin/arm64

### LINKED ISSUE #136: Builders for continuous integration
The combination of TravisCI (Linux/OS X) and AppVeyor (Windows) is not enough.

v1.3.0 was patched for linux/arm64, but we have no CI for arm64 to ensure that it keeps on working. We need a more comprehensive solution that covers arm64, ppc, and Solaris.

The best bet is to look into the Builders used by Go itself.
https://github.com/golang/go/wiki/DashboardBuilders

Perhaps one option would be to ask the Go Team if fsnotify could become a golang.org/x/fsnotify subrepository, but maybe that isn't necessary.

Of note, there still could be [API changes](https://github.com/fsnotify/fsnotify/issues?q=is%3Aopen+is%3Aissue+label%3AAPI) that warrant a v2.0 of fsnotify.

arp242:
Use CircleCI to add CI for this.

Fixes #136

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,80 @@
+version: 2.1
+workflows:
+  main:
+    jobs: ['linux-arm64', 'ios']
+
+jobs:
+  # linux/arm64
+  linux-arm64:
+    machine:
+      image: ubuntu-2204:2022.07.1
+    resource_class: arm.medium
+    working_directory: ~/repo
+    steps:
+      - checkout
+
+      - run:
+          name: install-go
+          command: |
+            sudo apt -y install golang
+
+      - run:
+          name: test
+          command: |
+            uname -a
+            go version
+            go test -race ./...
+
+  # iOS
+  ios:
+    macos:
+      xcode: 12.5.1
+    working_directory: ~/repo
+    steps:
+      - checkout
+
+      - run:
+          name: install-go
+          command: |
+            export HOMEBREW_NO_AUTO_UPDATE=1
+            brew install go
+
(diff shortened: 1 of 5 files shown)


sagikazarmark:
Do you have an ETA for the next release? This PR in particular would help remove some wrapper code around fsnotify from Viper to let it build on unsupported platforms.

arp242:
> Do you have an ETA for the next release? This PR in particular would help remove some wrapper code around fsnotify from Viper to let it build on unsupported platforms.

Not really; something on the order of months.

If the wrapper code already exists and works, then I don't think we really need to rush out a release?

arp242:
Also, looking at the workaround, it seems you added it in 2021 @sagikazarmark over here: https://github.com/spf13/viper/blob/master/watch_wasm.go

Don't be afraid to report these kind of things upstream (i.e. here); the fix is trivial, and it almost certainly would have been in the v1.6.0 release ... if I had known about it. I just happened to stumble on it in this PR, which is why it's fixed here.

Doing a release now with just some minor bugfixes just creates a lot of churn for everyone IMO. The workaround doesn't seem so bad to keep for the time being.

sagikazarmark:
No rush on removing the wrapper now. Just wanted to know if a release would happen in the near future. I would have kept the relevant PR open.

TBH I completely forgot about it. The only reason it came to my attention is because someone asked for adding AIX to the list in addition to JS.

Thanks for all the work you are doing on fsnotify!

PR #530, opened 2022-10-20
Check privileges for symlink tests on Windows

mattn:
Tests for symlink always fail when run the tests as non-admin users on Windows since Windows require privileges to create symlinks. This CL change to check the privileges.

(Can I put label `hacktoberfest-accepted` on this PR ?)

### DIFF
--- a/fsnotify_test.go
+++ b/fsnotify_test.go
@@ -166,6 +166,9 @@ func TestWatch(t *testing.T) {
 				// behaviour too.
 				t.Skip("broken on macOS")
 			}
+			if !hasPrivilegesForSymlink() {
+				t.Skip("does not have privileges for symlink on this OS")
+			}
 
 			file := join(tmp, "file")
 			link := join(tmp, "link")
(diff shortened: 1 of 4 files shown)


fsnotify_test.go
```diff
@@ -972,7 +989,8 @@ func TestAdd(t *testing.T) {
 }
 
 // TODO: should also check internal state is correct/cleaned up; e.g. no
-//       left-over file descriptors or whatnot.
+//
+//	left-over file descriptors or whatnot.
```
mattn, line 993: oop. I'll revert this part and some below.

arp242:
Can you put the `helper_.go` files in the `internal` package? There's already `internal/windows.go` and `internal/unix.go`, although the unix.go one has odd build tags; maybe add a unix2.go.

Just keeps the root a bit cleaner of files 😅

mattn:
Thanks.

mattn:
Ah, newer Go fmt breaks indentation.

arp242:
> Ah, newer Go fmt breaks indentation.

I guess I still use an older `goimports` that I haven't updated since 1.19. Bit annoying it reformats it; I like to indent my TODO comments like that, but can remove the indentation if it's causing gofmt problems.

mattn:
I've had too much coffee and I'm full. ☕😂

image

arp242 (approved):

arp242:
Yeah, the illumos tests always take forever, although never seen it run that long. I'll just merge it – thanks!

mattn:
Thank you.

PR #532, opened 2022-11-03
Remove TestInotifyNoBlockingSyscalls

### LINKED ISSUE #531: TestInotifyNoBlockingSyscalls fail on arm
**To Reproduce**
`GOARCH=arm go test ./...`

```
--- FAIL: TestInotifyNoBlockingSyscalls (6.02s)
    inotify_test.go:400: Got a nonzero diff 2. starting: 12. ending: 14
FAIL
```

**Expected behavior**
Not error

Linux: lsb_release -a

```
Distributor ID: Debian
Description:    Debian GNU/Linux bookworm/sid
Release:        n/a
Codename:       bookworm
```

arp242:
This test checks if no additional OS threads are created; this was added after the old epoll-based inotify code was removed (which would create additional threads). It's always been a bit flaky: it currently already runs the test twice and "fails" only if both fail.

We don't really control the Go scheduler and when it creates/destroys OS threads, or have all that much insight in it (the "threadcreate" profile in pprof has been broken for years). I tried to make this test a bit more reliable a few months ago, and the run-it-twice "solution" was the best I could come up.

It still causes some failures though; see the links below. I don't know why it consistently fails on armle (and not e.g. arm64), but that seems a Go thing, rather than a fsnotify thing.

I'm not sure if this test adds all that much value; I don't really foresee a regression where we accidentally create more OS threads than we should.

Fixes #531
Also see: https://bugs.launchpad.net/ubuntu/+source/golang-fsnotify/+bug/1971967

### DIFF
--- a/backend_inotify_test.go
+++ b/backend_inotify_test.go
@@ -5,7 +5,6 @@ package fsnotify
 
 import (
 	"errors"
-	"fmt"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


horahoradev (approved):
LGTM! If this test is causing us problems, yeah, let's just rip it out. The original issue has been fixed, and regression is unlikely.

PR #540, opened 2022-11-17
Add recursive watcher for Windows backend

### LINKED ISSUE #21: Subtree watch on Windows

arp242:
Recursive watches can be added by using a "/..." parameter, similar to the Go command:

	w.Add("dir")         // Behaves as before.
	w.Add("dir/...")     // Watch dir and all paths underneath it.

	w.Remove("dir")      // Remove the watch for dir and, if
	                     // recursive, all paths underneath it too

	w.Remove("dir/...")  // Behaves like just "dir" if the path was
	                     // recursive, error otherwise (probably
	                     // want to add recursive remove too at some
	                     // point).

The advantage of using "/..." vs. an option is that it can be easily specified in configuration files and the like; for example from a TOML file:

	[watches]
	dirs = ["/tmp/one", "/tmp/two/..."]

Options for this were previously discussed at:
https://github.com/fsnotify/fsnotify/pull/339#discussion_r788246013

This should be expanded to other backends too; I started with Windows because the implementation is the both the easiest and has the least amount of control (just setting a boolean parameter), and we can focus mostly on writing tests and documentation and the for it, and we can then match the inotify and kqueue behaviour to the Windows one.

Fixes #21

### DIFF
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -66,6 +66,16 @@ import (
 // The sysctl variables kern.maxfiles and kern.maxfilesperproc can be used to
 // control the maximum number of open files, as well as /etc/login.conf on BSD
 // systems.
+//
+// # Windows notes
+//
+// Paths can be added as "C:\path\to\dir", but forward slashes
+// ("C:/path/to/dir") will also work.
+//
+// The default buffer size is 64K, which is the largest value that is guaranteed
+// to work with SMB filesystems. If you have many events in quick succession
+// this may not be enough, and you will have to use [WithBufferSize] to increase
+// the value.
 type Watcher struct {
 	// Events sends the filesystem change events.
 	//
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -1,6 +1,9 @@
 //go:build solaris
 // +build solaris
 
+// Note: the documentation on the Watcher type and methods is generated from
+// mkdoc.zsh
+
 package fsnotify
 
 import (
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -270,8 +270,8 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 // Remove stops monitoring the path for changes.
 //
 // If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
-// entire recusrice watch will be removed. You can use both "/tmp/dir" and
-// "/tmp/dir/..." (they behave identical).
+// entire recursive watch will be removed. You can use either "/tmp/dir" or
+// "/tmp/dir/..." (they behave identically).
 //
 // You cannot remove individual files or subdirectories from recursive watches;
 // e.g. Add("/tmp/path/...") and then Remove("/tmp/path/sub") will fail.
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -92,6 +92,8 @@ type Watcher struct {
 	//                      you may get hundreds of Write events, so you
 	//                      probably want to wait until you've stopped receiving
 	//                      them (see the dedup example in cmd/fsnotify).
+	//                      Some systems may send Write event for directories
+	//                      when the directory content changes.
 	//
 	//   fsnotify.Chmod     Attributes were changed. On Linux this is also sent
 	//                      when a file is removed (or more accurately, when a
@@ -184,7 +186,7 @@ func (w *Watcher) Close() error {
 //
 // A path can only be watched once; attempting to watch it more than once will
 // return an error. Paths that do not yet exist on the filesystem cannot be
-// added.
+// watched.
 //
 // A watch will be automatically removed if the watched path is deleted or
 // renamed. The exception is the Windows backend, which doesn't remove the
@@ -200,8 +202,9 @@ func (w *Watcher) Close() error {
 // # Watching directories
 //
 // All files in a directory are monitored, including new files that are created
-// after the watcher is started. Subdirectories are not watched (i.e. it's
-// non-recursive).
+// after the watcher is started. By default subdirectories are not watched (i.e.
+// it's non-recursive), but if the path ends with "/..." all files and
+// subdirectories are watched too.
 //
 // # Watching files
 //
@@ -266,8 +269,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
+//
+// You cannot remove individual files or subdirectories from recursive watches;
+// e.g. Add("/tmp/path/...") and then Remove("/tmp/path/sub") will fail.
+//
+// For other watches directories are removed non-recursively. For example, if
+// you added "/tmp/dir" and "/tmp/dir/subdir" then you will need to remove both.
 //
 // Removing a path that has not yet been added returns [ErrNonExistentWatch].
 //
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -285,8 +285,8 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 // Remove stops monitoring the path for changes.
 //
 // If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
-// entire recusrice watch will be removed. You can use both "/tmp/dir" and
-// "/tmp/dir/..." (they behave identical).
+// entire recursive watch will be removed. You can use either "/tmp/dir" or
+// "/tmp/dir/..." (they behave identically).
 //
 // You cannot remove individual files or subdirectories from recursive watches;
 // e.g. Add("/tmp/path/...") and then Remove("/tmp/path/sub") will fail.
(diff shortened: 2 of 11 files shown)


mkdoc.zsh
```diff
@@ -166,6 +174,8 @@ events=$(<<EOF
 	//                      you may get hundreds of Write events, so you
 	//                      probably want to wait until you've stopped receiving
 	//                      them (see the dedup example in cmd/fsnotify).
+	//                      Some systems may send Write event for directories
+	//                      when the directory content changes.
```
arp242, line 178: This behaviour is kind of annoying, IMO; if you create a new file or directory it sends a Write on the directory, too:

	>  fsnotify watch ./test/...

	> md NEWDIR
	19:36:19.7176   1 CREATE        "test\\NEWDIR"

	> cd NEWDIR
	> md SUBDIR
	19:36:29.4840   2 CREATE        "test\\NEWDIR\\SUBDIR"
	19:36:30.4888   3 WRITE         "test\\NEWDIR"

	> cd SUBDIR
	> echo x >newfile
	19:36:56.9763   4 CREATE        "test\\NEWDIR\\SUBDIR\\newfile"
	19:36:56.9786   5 WRITE         "test\\NEWDIR\\SUBDIR\\newfile"
	19:36:58.2154   6 WRITE         "test\\NEWDIR\\SUBDIR"

	> cd ../..
	> echo asd >zxczxc
	19:38:40.8707   7 CREATE        "test\\zxczxc"
	19:38:40.8741   8 WRITE         "test\\zxczxc"

This is also the case with non-recursirve watches:

	> fsnotify watch ./test2

	> md asd
	19:45:37.6129   1 CREATE        "test2\\asd"

	> echo zxc > zxczxc
	19:45:42.9015   2 CREATE        "test2\\zxczxc"
	19:45:42.9045   3 WRITE         "test2\\zxczxc"

	> cd asd
	> echo zxc > zxczxc
	19:45:49.0316   4 WRITE         "test2\\asd"

It's also inconsistent: you don't get this Write event on the root directory.

The reason is because it seems to use timestamps:

> FILE_NOTIFY_CHANGE_LAST_WRITE
>
> Any change to the last write-time of files in the watched directory or subtree
> causes a change notification wait operation to return. The operating system
> detects a change to the last write-time only when the file is written to the
> disk. For operating systems that use extensive caching, detection occurs only
> when the cache is sufficiently flushed. 
>
> FILE_ACTION_MODIFIED
> 
> The file was modified. This can be a change in the time stamp or attributes. 

I don't see any way to change or disable this. So ... it is what it is, I guess :-/

arp242:
CC: @nicks @milas

horahoradev:
will review tomorrow if someone doesn't get to it first, thanks

nshalman:
I will try to find some time to review this as soon as I can.

backend_fen.go
```diff
@@ -266,8 +269,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
```
milas, line 274:
```suggestion
// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
// entire recursive watch will be removed. You can use either "/tmp/dir" or
// "/tmp/dir/..." (they behave identically).
```

backend_inotify.go
```diff
@@ -281,8 +284,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
```
milas, line 289:
```suggestion
// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
// entire recursive watch will be removed. You can use either "/tmp/dir" or
// "/tmp/dir/..." (they behave identically).
```

backend_other.go
```diff
@@ -162,8 +165,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error { return nil }
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
```
milas, line 170:
```suggestion
// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
// entire recursive watch will be removed. You can use either "/tmp/dir" or
// "/tmp/dir/..." (they behave identically).
```

backend_windows.go
```diff
@@ -258,8 +265,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
```
milas, line 270:
```suggestion
// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
// entire recursive watch will be removed. You can use either "/tmp/dir" or
// "/tmp/dir/..." (they behave identically).
```

backend_windows.go
```diff
@@ -1,6 +1,10 @@
 //go:build windows
 // +build windows
 
+// Windows backend based on ReadDirectoryChangesW()
```
milas, line 4: RE: Godoc, might be worth changing all instances of "`/some/path/...`" to be "`/some/path/...` or `X:\some\path\...` depending upon platform" or similar

backend_kqueue.go
```diff
@@ -288,8 +291,15 @@ func (w *Watcher) AddWith(name string, opts ...addOpt) error {
 
 // Remove stops monitoring the path for changes.
 //
-// Directories are always removed non-recursively. For example, if you added
-// /tmp/dir and /tmp/dir/subdir then you will need to remove both.
+// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
+// entire recusrice watch will be removed. You can use both "/tmp/dir" and
+// "/tmp/dir/..." (they behave identical).
```
milas, line 296:
```suggestion
// If the path was added as a recursive watch (e.g. as "/tmp/dir/...") then the
// entire recusrice watch will be removed. You can use either "/tmp/dir" or
// "/tmp/dir/..." (they behave identically).
```

backend_windows.go
```diff
@@ -507,6 +526,10 @@ func (w *Watcher) remWatch(pathname string) error {
 	watch := w.watches.get(ino)
 	w.mu.Unlock()
 
+	if recurse && !watch.recurse {
+		return fmt.Errorf("can't use /... with non-recursive watch %q", pathname)
```
milas, line 530: At the very least, I'd make this `\...` since it's a Windows-specific error

milas (approved):

milas:
(Oops - submitted review without comment.)

The reasoning for using `/...` seems fine to me & logic is sound. My suggestions were all pedantic grammatical / error-related & safe to ignore :upside_down_face:

backend_fen.go
arp242: I see you left comments on all of these; that's not really necessary: the documentation for the `Watcher` types are all generated from mkdoc.zsh (it's not obvious at all its generated, maybe I should add a comment at the top of the file). I'll update it from there. (Also: I just have been asleep when I wrote this...)

backend_windows.go
arp242: I added a note in the "Windows notes" section for this. I think using "/some/path/... or X:\some\path\... depending upon platform" on every case where a path is mentioned would get a bit long, and I don't really expect people to be confused by it. Besides, forward slashes work well on Windows, and have since at least Windows 2000 AFAIK so it's not even needed to use \ instead of /.

arp242:
I'll probably merge this in the weekend, so if someone still wants to do a review: now's your chance 😅

tigerinus:
@arp242 - this is recursive watcher still valid?

README.md says that:

```
you must add watches for any directory you want to watch (a recursive watcher is on the roadmap: https://github.com/fsnotify/fsnotify/issues/18)
```

PR #544, opened 2023-01-05
Idiomatic bitwise operations

diist:
No-op replacement of `x&y == y` by `x&y != 0` in `Op.Has`.


arp242:
`o&h == h` makes more sense to my brain. I don't think it's really worth fussing about which is "better" than the other.

arp242:
I did end up applying this patch after discussion in #551, FYI.

PR #552, opened 2023-01-30
Fix: Ordering of channel closure

tigh-latte:
We updated `fsnotify` to `1.6.0` internally there today and noticed some of our test cases were failing.

We listen for the closure of `Events` in a `select` statement to know when some file operation has been completed. However, in this `select` we also listen for the closure of `Errors` just incase.

In `v1.5.4` all closures were [handled in their own defer](https://github.com/fsnotify/fsnotify/blob/v1.5.4/inotify.go#L195-L196), in `v1.6.0` these closures are handled in a single defer. However this causes a change in behaviour that I am not sure is intentional. Defers are handled in reverse order so `Events` was closed before `Errors`, but now as the order of the closure wasn't flipped, so now signals are sent in reverse order.

So, `doneResp` returns (and thus `Close() error`) before all closing functionality has been completed.

This just flips the ordering back to how it was in previous versions.

### DIFF
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -452,9 +452,9 @@ func (w *Watcher) WatchList() []string {
 // received events into Event objects and sends them via the Events channel
 func (w *Watcher) readEvents() {
 	defer func() {
-		close(w.doneResp)
-		close(w.Errors)
 		close(w.Events)
+		close(w.Errors)
+		close(w.doneResp)
 	}()
 
 	var (
--- a/backend_inotify_test.go
+++ b/backend_inotify_test.go
@@ -134,3 +134,38 @@ func TestRemoveState(t *testing.T) {
 	}
 	check(0)
 }
+
+func TestInotifyCloseOrder(t *testing.T) {
+	t.Parallel()
+
+	var (
+		tmp = t.TempDir()
+	)
+
+	w := newWatcher(t, tmp)
+	addWatch(t, w, tmp)
+
+	c := make(chan struct{})
+
+	var closed bool
+	go func() {
+		select {
+		case _, ok := <-w.Events:
+			if !ok {
+				closed = true
+			}
+		case _, ok := <-w.Errors:
+			_ = ok
+		}
+		close(c)
+	}()
+
+	time.Sleep(1 * time.Second)
+	w.Close()
+
+	<-c
+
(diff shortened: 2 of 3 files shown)


arp242:
> noticed some of our test cases were failing

Do you have more details on the failing test cases? It would be great to add them here too, so we can be sure the behaviour is correct on all backends and that it doesn't regress in the future.

Tigh-Gherr:
Sure I'll get something written up here shortly

Tigh-Gherr:
Pretty much we a select like so:
```go
watch_loop:
	for {
		select {
			. . . // do stuff with every matching event
			if !ok {
				break watch_loop
			}
			. . . // handle error
			if !ok {
				break watch_loop
			}
	}
```

The problem this lib introduces is that `// do stuff with every matching event` has some code that is expected to be ran on closure because it means we have a guarantee that an operation is completed that would be otherwise unknowable (without writing very hacky code). Since errors is closing first, it never actually gets ran.

I could write a simple test case that has a select and flips a bool to true if `!ok` for `watcher.Events` if you like?

arp242:
> I could write a simple test case that has a select and flips a bool to true if !ok for watcher.Events if you like?

Sounds good; as long as it will reliably fail before these changes. I think it's important to have a reliable test for this as the behaviour is so subtle (arguably, structuring the code in way you have isn't the best way to do things, but breaking existing code isn't good so it should be fixed).

Can add as a new subtest under `TestClose()` in `fsnotify_test.go`. If a single test run doesn't reliably fail maybe adding it in a loop that runs 10 times might be an option.

It might be that it will fail on some other platforms in the CI; but we can fix that if it comes up.

Tigh-Gherr:
I can look into re-orging the code and maybe just not caring about errors closing, however, I do think that the closure of inotify is a valid hook to listen to so receiving it should be reliable.

I've added a v simple test case that just flips a bool on closure of events, it's reliable in my testing. Lmk what you think.

tigh-latte:
going to close this as it's a subtle change across versions that is hard to test, and it only seemed to break functionality for me and I've thus updated my code to accommodate.

arp242:
Cheers; I spent a bit of time trying to get the tests reliable a few months back but never really managed to get it right.

PR #554, opened 2023-02-01
Pin golang.org/x/sys

kylie-sre:
Pinning this module to a proper version as it was causing me issues due to the pretty old version of `x/sys` in `go.mod`. Tests are passing, vet is clean, and `fmt` is good.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,7 @@ module github.com/fsnotify/fsnotify
 
 go 1.17
 
-require golang.org/x/sys v0.0.0-20220908164124-27713097b956
+require golang.org/x/sys v0.4.0
 
 retract (
 	v1.5.3 // Published an incorrect branch accidentally https://github.com/fsnotify/fsnotify/issues/445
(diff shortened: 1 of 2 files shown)


arp242:
What problems are you having specifically? I'm not sure why changing the version should really fix anything 🤔

Updating the version would also change the minimum supported Go release. I'd prefer to keep it as low as possible unless there's a specific reason to upgrade.

arp242:
The commit after the current version updates the requirement to Go 1.17:
https://github.com/golang/sys/commit/aba9fc2a8ff2c9439446386f616b860442f0cf9a

Everything after that doesn't seem to update the minimally required Go version.

Ubuntu 22.04 (the most recent LTS) ships with Go 1.18, so that's fine. Debian 11 has Go 1.15; this already doesn't work as Go 1.16 is required.

Go 1.16 (released August 2021) became mandatory with v1.5.1 (August 2021). Go 1.17 was released just a few days before it was tagged. Some people did experience problems (https://github.com/fsnotify/fsnotify/pull/400), but overall the feedback was limited.

So based on this, I think it's probably safe to just follow the Go release policy: "Each major Go release is supported until there are two newer major releases."

PR #572, opened 2023-07-12
Add NewBufferedWatcher()

arp242:
Follow-up to #550


backend_fen.go
```diff
@@ -133,8 +133,21 @@ type Watcher struct {
...
+// kernels have de-duplication logic, which means less activity in userspace and
+// generally better performance. However there may be some cases where a very
+// large buffer can enable an application to keep up with a very large number of
+// events. You will always be better off increasing the kernel buffers over
+// adding a large userspace buffer, but if you can't control the kernel buffer
+// then a buffered watcher is a reasonable option.
+func NewBufferedWatcher(sz uint) (*Watcher, error) {
 	w := &Watcher{
-		Events:  make(chan Event),
+		Events:  make(chan Event, sz),
```
mattn, line 150: negative sz should be checked and return error?

mattn: Ah, sorry, sz is uint.

mattn (approved):

PR #579, opened 2023-09-08
macOS 13 runner is available

shogo82148:
https://github.blog/changelog/2023-04-24-github-actions-macos-13-is-now-available/


arp242:
This should replace macos-12, not add to it. GitHub is fairly restrictive in the ratelimit for macOS jobs and we use it for all the BSD, illumos, and old Linux runners, so the strategy is "test oldest and tests newest, and assume everything in between is fine". Otherwise all the jobs take absolutely forever to run and especially with the intermittent test failures it's a complete pain.

Anyway, I did this in #585.

PR #580, opened 2023-09-08
run test with stable Go versions

shogo82148:


arp242:
This will cause an additional runner to be spawned, increasing build times; we spawn a lot of jobs already.

I also like to manually specify the versions; while Go is very stable and shouldn't introduce regressions, I prefer to not have the CI shift underneath me.

PR #590, opened 2023-10-22
kqueue: ignore events with Ident=0

arp242:
On macOS it seems that sometimes an event with Ident=0 is delivered, and
no other flags/information beyond that, even though we never saw such a
file descriptor. For example in TestWatchSymlink/277 (usually at the
end, but sometimes sooner):

		# Okay, expected
		CREATE               "/foo"
		REMOVE               "/foo"
		CREATE               "/apple"
		CREATE               "/pear"
		RENAME               "/apple"
		REMOVE               "/pear"

		# One or more write events without any ident or information?!
		WRITE                ""

Printing some debug in the loop right after we receive events shows:

		unix.Kevent_t{Ident:0x2a, Filter:-4, Flags:0x25, Fflags:0x2, Data:0, Udata:(*uint8)(nil)}
		unix.Kevent_t{Ident:0x0,  Filter:-4, Flags:0x25, Fflags:0x2, Data:0, Udata:(*uint8)(nil)}

The first is a normal event, the second with Ident 0. No error flag, no
data, no ... nothing.

I read a bit through bsd/kern_event.c from the xnu source, but I don't
really see an obvious location where this is triggered – this doesn't
seem intentional, but idk...

Technically fd 0 is a valid descriptor, so only skip it if there's no
path, and if we're on macOS.

### DIFF
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sync"
 
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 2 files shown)


arp242:
To be honest I'm not entirely sure if this is actually the right thing to do ... Probably more testing is needed to determine exactly what's going on.

PR #604, opened 2024-03-02
Remove unnecessary code duplication

onee-only:
Modified if statement to avoid unwanted code duplication


onee-only:
I'm not sure if these test failures are caused by this PR. Should I look into it?

arp242:
Merged via  #612

PR #608, opened 2024-03-23
chore: fix some comments

majorteach:


arp242:
Please don't use bots to change things that don't really fix anything.

PR #617, opened 2024-04-24
kqueue: set O_CLOEXEC on close pipe fds

kdrag0n:
These file descriptors are being leaked to child processes because O_CLOEXEC isn't set.


arp242:
Thanks.

(CI failure is unrelated; fixed that in #618)

PR #620, opened 2024-04-25
inotify: don't send event for IN_DELETE_SELF when also watching the parent

arp242:
This would result in two events being sent.


zepatrik:
How is this not a breaking change? We used the second event to distinguish between files and directories being deleted, which we cannot do at all now.
https://github.com/ory/x/blob/3886ee392c7016bca5a8b0fd0ffef24604085171/watcherx/directory.go#L45-L66
Why is this not an option instead?

PR #622, opened 2024-04-26
kqueue: fix watching relative symlinks

arp242:
os.Readlink() just returns the value of the link.


arp242:
This needs more work because we don't store a link → target mapping.

PR #628, opened 2024-04-28
inotify, windows: track renames

### LINKED ISSUE #26: Improved "rename" support

arp242:
Add Event.RenamedFrom to track event renames; this is sent on a subsequent Create event; for example:

	Event{Op: Rename, Name: "/tmp/file"}
	Event{Op: Create, Name: "/tmp/renamed", RenamedFrom: "/tmp/file"}

For now this is unexported as renamedFrom because the kqueue and FEN backends are not yet implemented.

Updates #26

### DIFF
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -136,6 +136,24 @@ type Watcher struct {
 	done        chan struct{} // Channel for sending a "quit message" to the reader goroutine
 	doneMu      sync.Mutex
 	doneResp    chan struct{} // Channel to respond to Close
+
+	// Store rename cookies in an array, with the index wrapping to 0. Almost
+	// all of the time what we get is a MOVED_FROM to set the cookie and the
+	// next event inotify sends will be MOVED_TO to read it. However, this is
+	// not guaranteed – as described in inotify(7) – and we may get other events
+	// between the two MOVED_* events (including other MOVED_* ones).
+	//
+	// A second issue is that moving a file outside the watched directory will
+	// trigger a MOVED_FROM to set the cookie, but we never see the MOVED_TO to
+	// read and delete it. So just storing it in a map would slowly leak memory.
+	//
+	// Doing it like this gives us a simple fast LRU-cache that won't allocate.
+	// Ten items should be more than enough for our purpose, and a loop over
+	// such a short array is faster than a map access anyway (not that it hugely
+	// matters since we're talking about hundreds of ns at the most, but still).
+	cookies     [10]koekje
+	cookieIndex uint8
+	cookiesMu   sync.Mutex
 }
 
 type (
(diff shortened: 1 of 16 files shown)


arp242:
Do you know how to do with this illumos @nshalman? I spent some time looking at it, but I couldn't really figure out how to best do this.

Unless I missed something, there isn't really any support to track "file A got renamed to file B"?

Otherwise we can compare directory snapshots like with kqueue (still need to write this) I guess.

nshalman:
> Do you know how to do with this illumos @nshalman? I spent some time looking at it, but I couldn't really figure out how to best do this.
> 
> Unless I missed something, there isn't really any support to track "file A got renamed to file B"?
> 
> Otherwise we can compare directory snapshots like with kqueue (still need to write this) I guess.

This should certainly be possible somehow. https://illumos.org/man/3C/port_associate includes the text:

>        The following exception events are delivered when they occur. These event
>        types cannot be filtered.
> 
>          FILE_DELETE       /* Monitored file/directory was deleted */
>          FILE_RENAME_TO    /* Monitored file/directory was renamed */
>          FILE_RENAME_FROM  /* Monitored file/directory was renamed */
>          UNMOUNTED         /* Monitored file system got unmounted */
>          MOUNTEDOVER       /* Monitored file/directory was mounted over */

And we certainly have code handling these events... https://github.com/fsnotify/fsnotify/blob/main/backend_fen.go#L450-L470

I don't know what changes we would need. I'd have to take some time to look how you did this for inotify to see if I can figure out how to do something similar here. I imagine if we need to track additional information we might want to stash some other information in the cookie when we do our associations.

I'll try to find some time to look at this, but it will probably not be for at least a week.

Thanks for keeping me in the loop!

arp242:
I unexported `RenamedFrom` to `renamedFrom` for now, and the tests aren't run on platforms where this isn't supported yet (kq, fen). This way new features like this can be developed and merged in to main without holding up everything, and PRs can be a bit smaller in scope (I did some work on kqueue, which required rewriting significant bits of it).

To enable running the rename tests, add new platforms to `supportsRename()` in helpers_test.go over here: https://github.com/fsnotify/fsnotify/pull/628/files#diff-62f46d82044a129997b958c0f7eb6a95038395d337176418d3dffae98d9332d1R636

Issue #519, opened 2022-10-13
Support additional event types


arp242:
We only have Create, Remove, Rename, Write, and Chmod now, but many watchers support additional events. This has come up a few times:

- #24 - Extend, Link, Revoke
- #297 - Unmount
- #346 - Read
- #504 - Open, Close
- Probably some others?

Instead of creating a bunch of issues for every possible event, track them here.

All of this depends on #7; we don't want to spam people with e.g. Open and Close
events when they're not interested in them (which is most of the time), and some
events aren't supported on all platforms.

arp242:
I looked around a bit, and:

- Linux has OPEN and CLOSE events, FreeBSD has NOTE_OPEN, NOTE_CLOSE, NOTE_CLOSE_WRITE, but none of the other kqueue platforms do; Windows and FEN doesn't have it either.

- Unmount is available on inotify and FEN, but none of the kqueue platforms or Windows.

- Read is on FreeBSD, Linux (as ACCESS), FEN (as ACCESS), Windows (LAST_ACCESS), not none of the other kqueue platforms.

As far as I can see, there are no events that are shared by all platforms.

Need to think about how to best deal with this; #67 would be okay for some rare operations, but things like Read is available on quite a few platforms, and using a new .Sys() would be awkward.

One way might be creating a new UnportableOp type:

	// UnportableOp are file operations that are not available on all platforms.
	type UnportableOp uint32

	const (
		Read OP = 1 << iota
		Open
		Close
		// ...
	)

	type Event struct {
		Name       string
		Op         Op
		Unportable UnportableOp
	}

	func (e Event) HasUnportable(op UnportableOp) bool {
		return e.Unportable.Has(op)
	}


This way it's always clear you're using unportable operations and can't accidentally do `e.Has(fsnotify.Read)` and expecting it to work on all platforms; I don't trust people to read the documentation because experience has proven many don't (plus, it's easy to forget). On the other hand, having two types is a bit of a pain.

nickajacks1:
@arp242 What about leveraging GOOS build tags to limit which events can be used?
You could have a file or package for windows specific events, one for inotify specific events, etc
If each platform's non-portable bits had their own package, their usage limitations would be self documenting (eg `inotify.Close`, defined in github.com/fsnotify/fsnotify/pkg/inotify)
Would there be a problem with making this seamlessly handled by go build, without any kind of explicit opt-in?

[...]

PR #629, opened 2024-04-28
Support filtering events, add unsupported events

arp242:
This adds the ability to only listen for some event types. If you're only interested in Created events and you're getting a lot of Write events then you're just wasting CPU cycles

This also adds the ability listen on extra unportable event types; since this is so related I figured I might as well do both. Ideally we want to 1) make it very very obvious you're doing something unportable, and 2) make it reasonably easy "fallback" for platforms where this isn't supported.

Unportable events start with "Unportable", which should document their unportabilitiness. Also add a new Supports(Op) method, which should make adding fallback logic relatively painless.

For example, to use CloseWrite where supported, but falling back to Write when it's not:

	var op fsnotify.Op
	if w.Supports(fsnotify.UnportableCloseWrite) {
		op |= fsnotify.UnportableCloseWrite
	} else {
		op |= fsnotify.Create | fsnotify.Write
	}
	w.AddWith("/tmp", fsnotify.WithOp(op))

And then you can deal with this in the write loop. There's a full example in cmd/fsnotify/closewrite.go

TODO: need to write tests for this, update all platforms.

Updates #7
Updates #519

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -86,6 +86,31 @@ const (
 	// get triggered very frequently by some software. For example, Spotlight
 	// indexing on macOS, anti-virus software, backup software, etc.
 	Chmod
+
+	// File descriptor was opened.
+	//
+	// Only works on Linux and FreeBSD.
+	xUnportableOpen
+
+	// File was read from.
+	//
+	// Only works on Linux and FreeBSD.
+	xUnportableRead
+
+	// File opened for writing was closed.
+	//
+	// Only works on Linux and FreeBSD.
+	//
+	// The advantage of using this over Write is that it's more reliable than
+	// waiting for Write events to stop. It's also faster (if you're not
+	// listening to Write events): copying a file of a few GB can easily
+	// generate tens of thousands of Write events in a short span of time.
+	xUnportableCloseWrite
+
+	// File opened for reading was closed.
+	//
+	// Only works on Linux and FreeBSD.
+	xUnportableCloseRead
 )
 
 var (
@@ -105,6 +130,10 @@ var (
 	//  - windows:      The buffer size is too small; WithBufferSize() can be used to increase it.
 	//  - kqueue, fen:  Not used.
 	ErrEventOverflow = errors.New("fsnotify: queue or buffer overflow")
+
+	// ErrUnsupported is returned by AddWith() when WithOps() specified an
+	// Unportable event that's not supported on this platform.
+	xErrUnsupported = errors.New("fsnotify: not supported with this backend")
 )
 
 func (o Op) String() string {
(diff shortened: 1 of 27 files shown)


nickajacks1:
Nice, Supports was a great idea because it lets the programmer explicitly decide what the fallback behavior is. While I'd personally prefer compile time checks, this explicit runtime check method is probably more approachable for people and more in the spirit of fsnotify. A message warning about Unsupported* events in the Readme could also help to make sure devs are aware of any potential misuse.

arp242:
> While I'd personally prefer compile time checks, this explicit runtime check method is probably more approachable for people and more in the spirit of fsnotify. 

In principle I agree, however looking at this for a bit, one of my main issues with build tags is that authors will have to make two files like so:

has_closewrite.go:

    //go:build linux || freebsd

    var fsOp = fsnotify.UnportableCloseWrite

    func procEvent(e fsnotify.Event) {
        if e.Has(fsnotify.UnportableCloseWrite) {
            fmt.Println(e)
        }
    }

no_closewrite.go

    //go:build !linux && !freebsd

    var fsOp = fsnotify.Create | fsnotify.Write

    func procEvent(e fsnotify.Event) {
        // Fallback with timers as per ./cmd/fsnotify/dedup.go
    }

And this wouldn't be a huge deal, but those build tags aren't static, and platforms can add support for things like CloseWrite in the future, or fsnotify can add support for things that we didn't have before.

This is not hypothetical: NetBSD 10 (released last month) adds support for CloseWrite and a few other events. Maybe OpenBSD will add it next month or next year? Entirely possible. When/if we add support for fanotify, fsevents, polling, the feature sets for those backends will be different (but the GOOS build tag won't be).

So a 1:1 mapping of operating system to supported features is neither static nor perfect.

And it's not an uncommon situation for someone to be developing on, say, macOS, having a CI for Windows and Linux, and BSD systems will more or less "just work" automatically. By focusing on platforms (go:build linux) rather than features (w.Supports(fsnotify.UnportableCloseWrite)) things will probably break more on lesser-used systems, or use sub-optimal solutions on those systems.

All things considered, runtime checks are probably the right choice here, and prefixing the name with UnportableEvent should give enough of a clue that it's unportable, and people will have to make a conscious decision about this.

---

The main downside of Watcher.Supports is that it's tied to the Watcher. When I added AddWith() to pass options in #521, I imagined it could be used to select the backend, among other things; e.g.

    w.AddWith("/path", fsnotify.WithFanotify())
    w.AddWith("/path", fsnotify.WithPoll())

But that won't jibe with Watcher.Supports(). I'm okay with just telling people to use different Watcher instances for different backends, as that's probably quite a rare use case. But need to be able to override the default backend selection. Having a method for that is probably okay:

    func (w *Watcher) Backend(b Backend) error {
        if w.startedWatching {
            return errors.New("already started")
        }
        w.backend = b
        return nil
    }

Adding options to NewWatcher() was discussed in #550 when NewBufferedWatcher() was added.

arp242:
Alright, I think the basic API and implementation for inotify should be good. I unexported all the new methods/types and merged to main. After the other platforms have been implemented we can export things and offer a public API.

Merging it now means we don't have to keep a bunch of long-lived branches around but can implement things a bit more piece-meal.

fsnotify.go
```diff
@@ -184,6 +227,24 @@ func WithBufferSize(bytes int) addOpt {
...
+// or Chmod operations per second.
+//
+// This can also be used to add unportable operations not supported by all
+// platforms; unportable operations all start with "Unportable":
+// [UnportableOpen], [UnportableRead], [UnportableCloseWrite], and
+// [UnportableCloseRead].
+//
+// AddWith returns an error when using an unportable operation that's not
+// supported. Use [Watcher.Support] to check for support.
+func withOps(op Op) addOpt {
```
t33m, line 244: @arp242 , Hey, this is a really useful way to filter only the file events you need, but how this can be used if the function is private?

Issue #616, opened 2024-04-10
panic in v1.7.0


aakashbajaj:
### Describe the bug
Panic occured randomly in one of our microservices.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xe01628]

goroutine 327 [running]:
github.com/fsnotify/fsnotify.(*watches).remove(0xc0007bdec0, 0x90de68?)
        /go/pkg/mod/github.com/fsnotify/fsnotify@v1.7.0/backend_inotify.go:182 +0x88
github.com/fsnotify/fsnotify.(*Watcher).readEvents(0xc00012cd00)
        /go/pkg/mod/github.com/fsnotify/fsnotify@v1.7.0/backend_inotify.go:535 +0x4a5
created by github.com/fsnotify/fsnotify.NewBufferedWatcher
        /go/pkg/mod/github.com/fsnotify/fsnotify@v1.7.0/backend_inotify.go:270 +0x1f0
```

### Code to Reproduce

Not able to reproduce

### File operations to reproduce

NA

### Which operating system and version are you using?

Linux 6.1.58+ SMP PREEMPT_DYNAMIC Mon Jan 29 15:19:25 UTC 2024 x86_64 Linux

### Which fsnotify version are you using?

v1.7.0

### Did you try the latest main branch?

No

arp242:
I can't do anything with this; you'll need to either give more details on how to reproduce or track down the problem yourself.

aakashbajaj:
I apologise but we are not able to repro this at all. And there's no specific application state that caused this to occur.
fsnotify is an indirect dependency in the service

arp242:
I mean, there's no details here at all; not even how or where you're using it, or what kind of events are happening, or ... anything.

The panic is triggered on IN_DELETE_SELF: the watched path was deleted. This is handled (and tested), but it seems some specific sequence of events cause this error for some reason, but I'm not going to spend ages trying random stuff to find out what.

*Any* clue on usage patterns would be useful. Or if you want to know exactly, use `replace github.com/fsnotify/fsnotify => ./fsnotify` in go.mod and add a few debug printfs – maybe adding a FSNOTIFY_DEBUG flag would be nice for this.

ChristopherAparicio:
I have the same panic happening time to time on my micro-services.

The library is used indirectly by viper

```
[...]

PR #639, opened 2024-05-23
fix: accident nil pointer error when removing watch descriptor without map value checking in watches

gqcn:
We've got a nil pointer issue that caused by fsnotify v1.7.0, but it still exists in latest codes, so I post this pr to add additional existence check for value of watch descriptors map.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@
 
 /test/kqueue
 /test/a.out
+
+# ignore ide files from idea
+.idea/
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -14,8 +14,9 @@ import (
 	"time"
 	"unsafe"
 
-	"github.com/fsnotify/fsnotify/internal"
 	"golang.org/x/sys/unix"
+
+	"github.com/fsnotify/fsnotify/internal"
 )
 
 type inotify struct {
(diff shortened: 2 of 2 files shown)


arp242:
This is #616.

I'm not sure this is a good fix; ideally *should* happen is:

	func Add()    { lock(); do_stuff(); unlock() }
	func Remove() { lock(); do_stuff(); unlock() }

	func readEvents() {
		for {
			events := waitUntilDataFromInotify()
			lock()
			for event := range events {
				process(event)
			}
			unlock()
		}
	}

So once we've got events from inotify process them and only modify watched directories with Add() and Remove() after we've finished processing them.

But this is not as straight-forward as it sounds, because in the process() it will send to the Events and Errors channels, which may block if not read, in which case it will remain locked.

I'm not 100% sure that's an issue, but some tests fail on this.

gqcn:
> This is #616.
> 
> I'm not sure this is a good fix; ideally what _should_ happen is:
> 
> ```
> func Add()    { lock(); do_stuff(); unlock() }
> func Remove() { lock(); do_stuff(); unlock() }
> 
> func readEvents() {
> 	for {
> 		events := waitUntilDataFromInotify()
> 		lock()
> 		for event := range events {
> 			process(event)
> 		}
> 		unlock()
> 	}
> }
> ```
> 
> So once we've got events from inotify process them and only modify watched directories with Add() and Remove() after we've finished processing them.
> 
> But this is not as straight-forward as it sounds, because in the process() it will send to the Events and Errors channels, which may block if not read, in which case it will remain locked.
> 
> I'm not 100% sure that's an issue, but some tests fail on this.

Unit testing cases are failed accidentally in my project that depends on `fsnotify`, it seems the `remove` function was called multiple times with the same `wd`. I wonder if there's perfect fix for this, but this order-check resolves my issue.

gqcn:
@arp242  I think, whatever the root cause is, this existence-check would guarantee the robustness removing watch descriptors. How do you think of it? Can this pr be merged?

fals:
Hey, what is prevent us to merge this change? This became a recurring issue after updating to 1.7.0, we don't have a release of the library since Oct last year, would be nice moving forward with it.

arp242:
Fix in #650

PR #647, opened 2024-10-22
Fixed the AIX

ThakkarChandresh:
I've tried executing this aix branch in my aix machine but it had some problems executing it, which I've fixed, and tested all functionalities and everything is working


arp242:
I have no way of testing this.

I just removed the aix branch here. I have no interest in going out of my way to support a $7.5 billion net income company in my spare time that can't even be bothered to respond to my requests to make some testing environment available.

PR #651, opened 2024-10-28
Expose and document Supports, Unportable*

arp242:
Only implemented on Linux/inotify for now. I had hoped to finish some more platforms before exporting this, but I accidentally deleted my work on that a few months ago and haven't had the energy to restart it.

Lots of people want it, and there's are a few "forks" of fsnotify which export this. So just add it in the release.


arp242:
I thought I had done the filtering of the standard events (Create, Remove, etc.) for other backends, but I didn't. This can't really be merged until that's done, because you'll get different events on different platforms.

So that needs to be done first.

I wouldn't expect movement on this any time soon, at least, not from me. I had to spend time of my bank holiday last week to finish the 1.8 release.

Issue #656, opened 2024-11-26
Deadlock in sendEvent


bboreham:
### Describe the bug

After updating to fsnotify 1.8.0, the Prometheus project started seeing these goroutines reported at the end of tests, on the Windows platform:

E.g. https://github.com/prometheus/prometheus/actions/runs/12007123952/job/33467072512?pr=15449

```
Goroutine 470 in state select, locked to thread, with github.com/fsnotify/fsnotify.(*readDirChangesW).sendEvent on top of the stack:
github.com/fsnotify/fsnotify.(*readDirChangesW).sendEvent(0xc0006dbbc0, {0xc000038540?, 0xc0005d0450?}, {0x0?, 0xffffffff?}, 0x200?)
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/backend_windows.go:72 +0x176
github.com/fsnotify/fsnotify.(*readDirChangesW).readEvents(0xc0006dbbc0)
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/backend_windows.go:622 +0xdfd
created by github.com/fsnotify/fsnotify.newBufferedBackend in goroutine 710
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/backend_windows.go:55 +0x198

 Goroutine 710 in state chan receive, with github.com/fsnotify/fsnotify.(*readDirChangesW).AddWith on top of the stack:
github.com/fsnotify/fsnotify.(*readDirChangesW).AddWith(0xc0006dbbc0, {0xc0007aa400, 0x25}, {0x0, 0x0, 0x3b?})
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/backend_windows.go:141 +0x38d
github.com/fsnotify/fsnotify.(*readDirChangesW).Add(0xc0007aa400?, {0xc0007aa400?, 0xd0?})
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/backend_windows.go:111 +0x1f
github.com/fsnotify/fsnotify.(*Watcher).Add(...)
	C:/Users/runneradmin/go/pkg/mod/github.com/fsnotify/fsnotify@v1.8.0/fsnotify.go:313
[...]

PR #657, opened 2024-11-26
Make BufferedWatcher buffered again

bboreham:
This detail got lost in #632.

### DIFF
--- a/fsnotify.go
+++ b/fsnotify.go
@@ -266,7 +266,7 @@ func NewWatcher() (*Watcher, error) {
 // cases, and whenever possible you will be better off increasing the kernel
 // buffers instead of adding a large userspace buffer.
 func NewBufferedWatcher(sz uint) (*Watcher, error) {
-	ev, errs := make(chan Event), make(chan error)
+	ev, errs := make(chan Event, sz), make(chan error)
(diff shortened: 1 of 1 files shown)


bboreham:
It seems to me you are proposing a change to the design and/or API.  I didn't spend enough time on this to make such judgements.

I was reporting #656, and found a workaround which was defeated by this simpler problem.

However in the project where fsnotify is used, we pinned to the older version 1.7.

mweibel:
@bboreham maybe it wasn't clear what my intention was then:

- the <=1.7 code used a buffered channel of size 50 for windows in all cases (using `NewWatcher` or using `NewBufferedWatcher`)
- the 1.8 code uses a buffered channel only if using `NewBufferedWatcher`

This is because the creation of the `Events` channel (with/without buffer) is in the platform-agnostic `fsnotify.go` and only the backend creation is platform-specific.

Therefore, this fix here is valid for users using `NewBufferedWatcher` but doesn't fix users using `NewWatcher` on windows. This is a breaking change from the previous version and seems to deadlock even when merging this change.

bboreham:
> doesn't fix users using NewWatcher

Agreed. I did not claim this PR fixes any issue, especially not #656.

PR #668, opened 2025-01-20
test: remove test with hardcoded dates

### LINKED ISSUE #665: Tests fail in  year 2025
This test will not work in 2025

https://github.com/fsnotify/fsnotify/blob/a9bc2e01792f868516acf80817f7d7d7b3315409/internal/ztest/diff_test.go#L139-L150

mikelolasagasti:
The test would require to be changed each year, and previous test already test the behaviour.

Resolves #665


arp242:
Fixed by #675.

PR #675, opened 2025-03-05
Fix hardcoded year in test

lawmurray:
A tiny fix: the year 2024 is hard-coded in one of the tests, so that it fails in The Year of the Linux Desktop.


lawmurray:
Sorry, I just realised that the very last PR #668 addresses the same! Well, this one comes with a fix and a joke, so you can choose :laughing:

arp242:
(Note: CI failures are unrelated; looks like some stuff "bitrotted" and needs some fixing)

Issue #673, opened 2025-02-20
Any timeline that the recursive flag for addWatch can be exposed?


hu13:
hi, i want to perform recursive watch by default on windows.

but the api does not yet expose this option. 
https://github.com/fsnotify/fsnotify/blob/main/fsnotify.go#L481

any timeline for that? or why is this currently not possible?

apologize if this question has been asked many times.

thanks.

SearidangPa:
Currently, the code base has this [function](https://github.com/fsnotify/fsnotify/blob/a9bc2e01792f868516acf80817f7d7d7b3315409/fsnotify.go#L485) to determine if the user wants the watcher to be recursive. 

But because of the `filepath.Clean` path function, when i input something like `syncRoot\...`, `filepath.Clean()` would remove the trailing `\...`

lawmurray:
There is also an `enableRecurse` flag that needs to be set to enable the feature. If you're happy with a dirty hack, you can use `go:linkname` to set that as I did [here](https://github.com/rclone/rclone/pull/8258/files/c9e28b34142eb93052ac95df6e2a85f592686677#diff-b7cbce6ef2eaf813cc1ce43c436e9a673fcb1b7b830f370378aa9a96228f0172) for an rclone PR, but apparently `go:linkname` is [being deprecated](https://huizhou92.com/p/%23-golang-1.23-changes-to-/golinkname-and-what-it-means-for-developers/), making it doubly dirty. Once you've done that though, you can pass in paths with trailing `\...` to get recursive watchers on Linux and Windows.

Looking at the code, fsnotify has recursive watchers for Windows (system provided) and Linux (implemented in fsnotify using non-recursive watchers) but I suppose is in need of a contribution for Mac before the feature is complete.

I do wonder whether, rather than specific Linux and Mac recursive watchers, there could be just one recursive watcher implemented around fsnotify's own interface, rather than the lower-level inotify and FSEvents interfaces. That's basically what this PR for rclone is doing, and it works on both Linux and Mac. Perhaps that code could be adapted for upstream to fsnotify. See [here](https://github.com/lawmurray/rclone/blob/7787b7d54155cabf58ec40a9fa0ccd047cd2f393/backend/local/changenotify_other.go#L136) for some further progress that's a little simpler than the PR linked above.

As a stopgap, perhaps the maintainers would consider a `SetEnableRecurse()` function to set the flag instead, so that at least the Linux and Windows recursive watchers become available.

[...]

PR #676, opened 2025-03-07
Add recursive backend that wraps any other non-recursive backend.

lawmurray:
Consider this a draft. It would be useful to hear whether there is interest in this approach before investing more time in it, and if so to get some help from others to debug platform specifics, especially for BSD, as I have little experience there. The public interface needs some consideration, currently I've just set this up to force the use of the recursive backend in all cases, as that was easiest for testing.

There is interest in supporting recursive watchers (e.g. #673, #114, #18). I have been interested in this for a PR on rclone also (rclone/rclone#8258), where I wrote a user-level recursive watcher around fsnotify that works on Linux and Mac, so in some sense this is looking at the feasibility of upstreaming that.

The idea of this approach is to provide a recursive backend that wraps any other non-recursive backend to provide support for recursive watchers, so that it is not necessary to add support for every backend separately. It does that using a pipeline approach, starting a goroutine that consumes from the existing event channel, does its work for recursion, and emits to a second event channel from which the user of fsnotify consumes.

For the Windows backend, where there is system-level support, it mostly passes through, although still acts as a wrapper to provide consistent behavior with create events. I realise the inotify backend has a user-level recursive implementation too; much of this is based on that actually, and might be seen as factoring that out to be usable by other backends too.

### DIFF
--- a/backend_fen_test.go
+++ b/backend_fen_test.go
@@ -23,21 +23,21 @@ func TestRemoveState(t *testing.T) {
 
 	check := func(wantDirs, wantFiles int) {
 		t.Helper()
-		if len(w.b.(*fen).watches) != wantFiles {
+		if len(w.b.(*recursive).b.(*fen).watches) != wantFiles {
 			var d []string
-			for k, v := range w.b.(*fen).watches {
+			for k, v := range w.b.(*recursive).b.(*fen).watches {
 				d = append(d, fmt.Sprintf("%#v = %#v", k, v))
 			}
 			t.Errorf("unexpected number of entries in w.watches (have %d, want %d):\n%v",
-				len(w.b.(*fen).watches), wantFiles, strings.Join(d, "\n"))
+				len(w.b.(*recursive).b.(*fen).watches), wantFiles, strings.Join(d, "\n"))
 		}
-		if len(w.b.(*fen).dirs) != wantDirs {
+		if len(w.b.(*recursive).b.(*fen).dirs) != wantDirs {
 			var d []string
-			for k, v := range w.b.(*fen).dirs {
+			for k, v := range w.b.(*recursive).b.(*fen).dirs {
 				d = append(d, fmt.Sprintf("%#v = %#v", k, v))
 			}
 			t.Errorf("unexpected number of entries in w.dirs (have %d, want %d):\n%v",
-				len(w.b.(*fen).dirs), wantDirs, strings.Join(d, "\n"))
+				len(w.b.(*recursive).b.(*fen).dirs), wantDirs, strings.Join(d, "\n"))
 		}
 	}
 
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -99,6 +99,13 @@ func (w *watches) remove(wd uint32) {
 	delete(w.wd, wd)
 }
 
+func (w *watches) existsPath(path string) bool {
+	w.mu.RLock()
+	defer w.mu.RUnlock()
+	_, ok := w.path[path]
+	return ok
+}
+
 func (w *watches) removePath(path string) ([]uint32, error) {
 	w.mu.Lock()
 	defer w.mu.Unlock()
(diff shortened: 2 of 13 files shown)


arp242:
Recursion already works for Windows and inotify; adding to it kqueue isn't actually that hard and I did it already last year, but then I accidentally deleted all my work by freeing up space and deleting my FreeBSD VM 😭

I need to re-do that work and rewrite the kqueue backend (again), not just for the recursion, but also for other features like event filtering and tracking renames.

That leaves us with illumos/FEN. I think adding it to that isn't too much effort, but I haven't looked at it yet. I'm also kind of okay shipping without FEN support if that's really a hang-up, because it's so little used. But we'll see how it goes.

So with that, it kind of makes this approach redundant? It's also something you can do from the "outside": that is, I don't think you strictly need any code in here to make it work?

lawmurray:
The idea is to factor out recursion in the design to ensure consistent behavior across all backends (that do not natively support it, so all but Windows I think?). It allows the inotify code to be simplified by removing the recursive bit, and there would be no need to finish recursive implementations for the other backends. Whereas, implementing recursion for each backend (that does not natively support it) separately may mean inconsistent behavior, more opportunity for bugs, and more code to maintain.

Of course, this is not my decision to make, just making the case for consideration.

arp242:
> The idea is to factor out recursion in the design to ensure consistent behavior across all backends (that do not natively support it, so all but Windows I think?). It allows the inotify code to be simplified by removing the recursive bit, and there would be no need to finish recursive implementations for the other backends

Yeah, that would be nice. I guess my concern is stuff like:

    if inEvent.Mask&unix.IN_MOVE_SELF == unix.IN_MOVE_SELF {
        if watch.recurse { // Do nothing
            return Event{}, true
        }

Which is not necessarily easy to reliably reproduce with a wrapper? There is some other stuff as well, such as renames within the watched directory: since the file descriptor/watch is retained, all we need to do is update the internal structure.

Other than that, adding recursion isn't that difficult as such. If you look at the ionotify backend then it's very little code and all of it is very straight-forward. It's just that other stuff such as bugs have always taken priority, and often it's quite time-consuming to debug/fix these in a cross-platform way, and that the general design of things is a bit messy in parts due to a long history (hence the need for a rewrite).

valoq:
Offering help with testing this feature. Hoping for it to make the next release

firetest770:
This feature would solve a few problem for me and I'm quite looking forward to this making it into a future release.
While I'm just a student, I would like to sponsor this feature in particular.
Is there any realistic chance for this to be implemented this year?

MK796:
I would like to add a concrete downstream use case and suggest giving recursive watch support higher priority.

Traefik uses fsnotify as the foundation of its file provider. Traefik currently loads configuration files recursively from the configured `providers.file.directory`, but with `watch: true` it only registers watches for the configured root and supported top-level configuration files. Changes deeper in the directory tree therefore do not reliably trigger a reload.

For Traefik, the preferred downstream behavior would be automatic recursive watching below the existing configured directory whenever `watch: true` is enabled.

The existing configured `directory` would remain the root of that recursive watch. Existing descendants and directories created after startup would be watched consistently. No additional recursion-specific user-facing option would be required.

This does not mean that fsnotify should make every existing watch recursive by default. fsnotify should expose a stable explicit recursive API, which Traefik could select internally for its existing `watch: true` behavior.

For Traefik's current release targets, recursive behavior needs to be consistent across:

- Windows
- Linux/inotify
- macOS, FreeBSD, and OpenBSD/kqueue

For fsnotify as a library, the public recursive API should also define the behavior of illumos/FEN for as long as that backend remains supported.

There is another implementation path for the same feature that should be cross-referenced here:

- #733 originally attempted to complete recursive watch support across all backends.
- #750 covered the remaining inotify fixes and was closed because commit 853adeb already addressed them.
- #751 adds recursive FEN support.
- #752 adds recursive kqueue support.
- #753 publicly enables recursive watches through the `/...` path suffix after the backend work is complete.

The common-wrapper approach proposed in this PR and the per-backend path represented by #751 and #752, followed by public enablement in #753, appear to be alternative implementation approaches rather than a linear dependency chain.

It would be useful to cross-link these efforts and clarify which direction the maintainers prefer. Downstream consumers need one stable and consistent cross-platform API.

The related Traefik history is:

https://github.com/traefik/traefik/issues/3986
https://github.com/traefik/traefik/pull/11117

Traefik declined its previous implementation partly because recursive behavior could differ between directories present at startup and directories created later. A stable fsnotify implementation with defined semantics would directly address that concern.

I have also been testing a downstream Traefik proof of concept across Linux, Windows, and macOS. These tests exposed a kqueue event-ordering race when a downstream implementation dynamically registers newly created paths. I can provide the reproduction and stress tests if useful.

PR #691, opened 2025-04-18
bug:fix lost rename event

ningmingxiao:
```
import os
 
def main():
    os.rename("/home/nmx/nmx.conflist", "/home/nmx/ansible/nmx.conflist")


if __name__ == '__main__':
    main()
```

use demo

```
package main

import (
    "log"

)

func main() {
    // Create new watcher.
    if err != nil {
        log.Fatal(err)
    }
    defer watcher.Close()

    // Start listening for events.
    go func() {
        for {
            select {
                if !ok {
                    return
                }
                log.Println("event:", event)
                if event.Has(fsnotify.Write) {
                }
                if !ok {
                    return
                }
                log.Println("error:", err)
            }
        }
    }()

    // Add a path.
    err = watcher.Add("/home/nmx/ansible/")
    if err != nil {
        log.Fatal(err)
    }

    // Block main goroutine forever.
    <-make(chan struct{})
}
```
will only print create log 

```
2025/04/18 14:17:58 event: CREATE        "/home/nmx/ansible/nmx.conflist"
```

### DIFF
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -203,7 +203,7 @@ func (w *inotify) AddWith(path string, opts ...addOpt) error {
 			flags |= unix.IN_DELETE | unix.IN_DELETE_SELF
 		}
 		if with.op.Has(Rename) {
-			flags |= unix.IN_MOVED_TO | unix.IN_MOVED_FROM | unix.IN_MOVE_SELF
+			flags |= unix.IN_MOVED_TO | unix.IN_MOVED_FROM | unix.IN_MOVE_SELF | unix.SYS_RENAME
 		}
 		if with.op.Has(Chmod) {
 			flags |= unix.IN_ATTRIB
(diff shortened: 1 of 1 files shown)


ningmingxiao:
@arp242

arp242:
I think you just need to watch /home/nmx as well?

Why would adding unix.SYS_RENAME fix anything? it's a inotify constant and this doesn't even compile?

ningmingxiao:
I use  ansible web -m copy -a "src=/etc/cni/net.d/nmx.conflist dest=/etc/cni/net.d/nmx.conflist" will generate random dir  src : /root/.ansible/tmp/ansible-tmp-1744869341.1990976-761162-201998177781176/source  see https://github.com/containerd/containerd/issues/11715 @arp242

ningmingxiao:
do  you have another way to fix ? @arp242

ningmingxiao:
after my pr I can catch rename event.

arp242:
Like I said, you need to watch both the source and destination. I don't know how that fits in containerd or ansible, but it's not a problem in fsnotify as near as I can tell.

PR #700, opened 2025-06-22
Update README example to use more idiomatic "block forever"

lesiw:
A tiny change to reflect current Go idioms.


arp242:
Tanks, but I prefer just using a channel.

PR #701, opened 2025-07-03
update to Go 1.24, add linter rules and fix issues

miparnisari:
Goal 1: improve developer experience

- Add a Makefile to run the basic commands without having to remember all the parameters
- Enforce stricter linting rules to catch more bugs before they get released

Goal 2: update Go and dependencies


miparnisari:
@arp242 could you review?

arp242:
Please don't submit spam like this.

arp242:
Well, making a bunch of highly opinionated changes that don't really fix anything with no disregard as to *why* some things are as they are certainly isn't helpful. Go 1.19 is supported for a reason. It's using that specific x/sys for a reason. The -race isn't in those specific tests for a reason. I don't like these Makefile wrappers. staticcheck is enough and other "linters" add little IMHO. Hostly, I have no idea why anyone would send these types of PRs.

Issue #353, opened 2020-11-18
AIX/ppc64 support


aklyachkin:
### Which operating system (GOOS) and version are you using?

AIX 7.2

### Please describe the issue that occurred.

Compiling different packages on AIX breaks because there are no fsnotify primitives for AIX.

### Are you able to reproduce the issue? Please provide steps to reproduce and a code sample if possible.

The easiest way to solve the problem is just to add the stub for AIX:

```
// ahafs.go
// Copyright 2010 The Go Authors. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.

// +build aix

package fsnotify

import (
	"errors"
)

// Watcher watches a set of files, delivering events to a channel.
type Watcher struct {
	Events chan Event
	Errors chan error
}

// NewWatcher establishes a new watcher with the underlying OS and begins waiting for events.
func NewWatcher() (*Watcher, error) {
	return nil, errors.New("AHAFS based watcher not yet supported for fsnotify\n")
}

// Close removes all watches and closes the events channel.
func (w *Watcher) Close() error {
	return nil
}

// Add starts watching the named file or directory (non-recursively).
func (w *Watcher) Add(name string) error {
	return nil
}

// Remove stops watching the the named file or directory (non-recursively).
func (w *Watcher) Remove(name string) error {
	return nil
}
```

nathany:
#314

aklyachkin:
Thank you! AIX has such functionality - see https://github.com/power-devops/ahafs

arp242:
Ah right. That could also be added here, but it's not something I can do as I don't have access to an AIX system. So someone else will have to pick that up and create a patch.

leoncamel:
@aklyachkin Any plan to integrate [aklyachkin](https://github.com/power-devops/ahafs) with fsnotify? Since some package may be pending on this feature on AIX.

aklyachkin:
@leoncamel In my lab I integrated them, but till now I didn't get it working stable. Sometimes it works, sometimes it doesn't. I didn't find the reason for it.

arp242:
The polling watcher should work on AIX regardless; I hope it'll get added relatively soon (still in 2022, maybe early 2023, although I promise nothing).

[...]

PR #702, opened 2025-07-09
Port fsnotify to AIX

KamathForAIX:
AIX is a UNIX-based operating system in Power Hardware. We have been porting Beats for AIX so that our users can use Auditbeat to track file operations, mainly create, remove, move, and rename.

This pull request ports fsnotify to AIX since this is a module needed by Auditbeat.

Currently we use the polling method to identify events.

Also, I wrote a simple test case for the same.

Kindly give me feedback so we can merge this pull request.

### DIFF
--- a/backend_aix.go
+++ b/backend_aix.go
@@ -1,0 +1,371 @@
+//go:build aix
+
+package fsnotify
+
+import (
+		"errors"
+		"io/ioutil"
+		"os"
+		"path/filepath"
+		"sync"
+		"time"
+)
+
+const (
+		sleepTime time.Duration = 100 * time.Millisecond
+)
+
+type aix_watcher struct {
+		*shared
+		Events chan Event
+		Errors chan error
+		mu       *sync.Mutex // Map access
+		isClosed bool
+		watches map[string]bool        // whatched files (recursive or not)
+		files   map[string]os.FileInfo // map of files.
+		maxEvents int
+		close     chan struct{}
+		done   chan struct{}
+}
+
+var defaultBufferSize = 0
+
+func newBackend(ev chan Event, errs chan error) (backend, error) {
+        w := &aix_watcher {
+                shared:  newShared(ev, errs),
+                Events:  ev,
+                Errors:  errs,
+                close:   make(chan struct{}),
+                mu:      new(sync.Mutex),
+                files:   make(map[string]os.FileInfo),
+                watches: make(map[string]bool),
+        }
+        go w.readEvents()
+        return w, nil
+}
+
+func (w *aix_watcher) listRecursive(name string) (map[string]os.FileInfo, error) {
+        fileList := make(map[string]os.FileInfo)
+
+        return fileList, filepath.Walk(name, func(path string, info os.FileInfo, err error) error {
+                if err != nil {
+                        return err
+                }
+                // Add the path and it's info to the file list.
+                fileList[path] = info
+                return nil
+        })
+}
+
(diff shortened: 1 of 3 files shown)


arp242:
My concerns with this are the lack of any testing/dev environment. Especially for long-term maintainership this is an issue: next year I may want to add $feature, and who will do the AIX bits for that? Someone reports a bug on AIX, who will fix that? You might say "yes, I'll do that" now and I fully believe that your intent is genuine, but in two years you might have changed jobs, or Auditbeat's business priorities changed, or whatever, and it will be left up to me.

Even with a testing environment it adds a huge burden. Without a dev environment it's basically impossible.

arp242:
By the way, to have AIX support you can do something like:

fsnotify_other.go

```
//go:build !aix

package watcher

type Watcher struct {
    fsnotify.Watcher
}
```

fsnotify_aix.go

```
//go:build aix

package watcher

type Watcher struct {
    Events chan Event
    Errors chan error
    mu     *sync.Mutex
    // ...
}

func (w *Watcher) Add(name string) error {
   ...
}
```

---

A detailed review would be a waste of both our times at this point as the lack of a dev environment makes this PR a non-starter, but as a general comment after skimming your code: you need to `gofmt` things, as well as run at least `staticcheck ./...` and fix the errors. I would also be surprised if `go test` passes. There are some other fairly basic issues as well (e.g. AddWith() just returns nil?), but fixing some of that would be a good start in any case.

KamathForAIX:
Hi @arp242 

Thank you for your reply.

> My concerns with this is the lack of any testing/dev environment. Especially for long-term maintainership this is a show-stopper: next year I may want to add $feature, and who will do the AIX bits for that? Someone reports a bug on AIX, who will fix that? You might say "yes, I'll do that" now and I fully believe that your intent is genuine, but in two years you might have changed jobs, or Auditbeat's business priorities changed, or something else, and it will be left up to me.
> 
> Even with a testing environment it adds a huge burden. Without a dev environment it's basically impossible. Any AIX support is a complete non-starter until that's resolved: either via some long-term access to a dev environment, or via some sort of credible long-term maintainership by someone who has access to one.
> 
> People have asked about AIX before (#353), including some IBM people, and I asked about a dev environment. They never even got back to me.

I have talked about the dev environment via OSU labs for AIX LPARs for open source communities within IBM. The VMs are getting set up, and I have been told it will take time before open-source communities get access to them. I understand that AIX isn't officially supported due to a lack of development environments. The idea was to submit this PR in such a way that it does not affect any other target, and we are committed to maintaining this on AIX. We are also happy to address issues and support AIX issues. 

We do something similar in the Meson community, where we do actively support AIX users, and maintainers tag us whenever an AIX issue is raised. You can always tag me or @ayappanec for any issues. This would benefit AIX users who want to use this package. Kindly let us know if you will consider.

> A detailed review would be a waste of both our times at this point as the lack of a dev environment makes this PR a non-starter, but as a general comment after skimming your code: you need to `gofmt` things, as well as run at least `staticcheck ./...` and fix the errors. I would also be surprised if `go test` passes. There are some other fairly basic issues as well (e.g. AddWith() just returns nil?), but fixing some of that would be a good start in any case.

Thank you for the review at a high level. Yes, everything does not pass as of now, and I want to work with you to resolve things one by one. I will address those two comments as of now and squash commits/push the code. I did not implement AddWith() as of now, but I can do the same. Will make these changes and get back with the test results once I do the same. Looks like this can fix some of the tests.

PR #705, opened 2025-08-11
Revert "Remove gccgo from CI"

purpleidea:
This reverts commit https://github.com/purpleidea/fsnotify/commit/a10c87150a06b15edb4c9f016147b13bbab77902.

I think gccgo is quite useful, and it's completely not dead. There's no reason this library needs generics.


arp242:
> There's no reason this library needs generics.

It was the usage of slices.Contains() or the like that triggered this change.

That didn't end up in the main branch in the end, and I suppose that theoretically this could be added again, but on the first gccgo compile error I'll just end up removing it from the CI again. So seems rather pointless.

PR #708, opened 2025-08-14
Remove explicit sysctl fs.inotify.max_user_watches hardcoding suggestion

ajoga:
Since Linux 5.11-rc1, fs.inotify.max_user_watches is dynamically computed up to 1048576 with regards to the addressable physical memory: https://github.com/torvalds/linux/commit/92890123749bafc317bbfacbe0a62ce08d78efb7 .


arp242:
Thanks; I think removing the "hard-coded" value is a bit too much, because you don't really have any option other than to "hard-code" it when you run out. The value of 124,983 came from my 8G laptop, and my 32G laptop has 243,992 as a default. In both cases I can see how one can run out of them by monitoring a (very) large directory. There are more files in my /home/martin for example.

So I just removed the reference to Linux 5.18 and added "The default values differ per distro
and available memory".

PR #725, opened 2025-12-12
[FIX] Kqueue race condition on macOS

Hersy:
Made possible through code reproduction example provided on issue #717
It allowed me to code a test that reproduces error when ran on macOS only, it also includes argument to run the test on Linux with Docker. I verrified test consistantly fails on macOS while succeeding on Linux.

When a file is deleted and quickly recreated with new content, the kqueue
backend could miss the WRITE event because the watch was set up after
sending the CREATE event. By the time the watch was established, the write
had already occurred.

Fixed by setting up the watch BEFORE sending the CREATE event, ensuring any
writes that happen after CREATE are captured.


mattn:
Please rebase onto main first — there are conflicts to resolve. Also the test has issues (single iteration with a 500ms timeout won't reliably catch the race; align with the existing `touch`/`rm`/`mkdirAll` helpers). Fix both and I'll take another look.

MikeeI:
I reproduced the ordering race on the same platform family as the report and found that the production change works, but the added test does not distinguish the fixed and unfixed implementations.

Environment: Darwin 24.6.0 arm64, Go 1.26.0.

Results:

- PR test on this branch: `go test -run TestDeleteRecreateRace -count=100` passes.
- The same PR test with the pre-fix `backend_kqueue.go` from merge base `bd7a7514a0fc6d187173c9f316267e722a96ec2c`: `-count=1000` also passes.
- A deterministic ordering test fails against that pre-fix backend on its first iteration: `timeout waiting for WRITE`.
- The same ordering test passes against this PR: `go test -run TestIssue717CreateThenWrite -count=10` passes (100 create/write cycles per run).

The distinguishing sequence is:

1. Watch a temporary directory.
2. Create a new exact-path file but do not write content yet.
3. Wait for that file's exact-path `CREATE` event.
4. Write and sync the still-open file.
5. Require a subsequent exact-path `WRITE` event under a bounded deadline.

That sequence makes the contract under review explicit. On the old implementation, `CREATE` is published before `internalWatch`, so the write performed by the receiver can occur before the file watch exists and no `WRITE` follows. With this PR, the file watch exists before `CREATE` is published, and the subsequent write is observed.

The current test instead performs `Remove` and `WriteFile` before consuming any event, then accepts any event once `ReadFile` sees the final content. A queued `REMOVE` or `CREATE` can therefore satisfy it after the write has already completed, even when the actual `WRITE` event was missed. It also does not close the watcher returned by `newWatcher`; repeated `-count` runs retain the reader goroutine and descriptors. I recommend replacing it with the ordered exact-path test above and registering `w.Close` with cleanup.

### Disclosure

Investigated thoroughly with openai-codex/gpt-5.6-sol (medium reasoning effort), using [Oh My Pi](https://github.com/can1357/oh-my-pi) as the agent framework.

This report is not generic or unreviewed AI-generated output. Its claims were checked against the cited evidence, and it includes the relevant detail intended to help maintainers resolve the issue.

If reports like this are not useful to the project, please let me know and I will refrain from submitting similar ones. My intent is to help without wasting maintainer time or energy or discouraging their work.

Thank you for your work.

PR #729, opened 2026-02-16
kqueue: allow watching directories only

WhyNotHugo:
The kqueue backend implicitly watches all files, requiring 1+N file descriptors. For usages which only care about new, deleted or moved files, this requires a huge amount of unnecessary file descriptors.

Add a flag to watch directories only, ignoring Write (mutation) events for files, but delivering events for new or deleted files.

### DIFF
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -24,6 +24,7 @@ type kqueue struct {
 	kq        int    // File descriptor (as returned by the kqueue() syscall).
 	closepipe [2]int // Pipe used for closing kq.
 	watches   *watches
+	dirsOnly  bool
 }
 
 type (
@@ -266,6 +267,10 @@ func (w *kqueue) AddWith(name string, opts ...addOpt) error {
 	with := getOptions(opts...)
 	if !w.xSupports(with.op) {
 		return fmt.Errorf("%w: %s", xErrUnsupported, with.op)
+	}
+
+	if with.dirsOnly {
+		w.dirsOnly = true
 	}
 
 	_, err := w.addWatch(name, noteAllEvents, false)
@@ -656,6 +661,10 @@ func (w *kqueue) internalWatch(name string, fi os.FileInfo) (string, error) {
 		// the flags used if currently watching subdirectory
 		info, _ := w.watches.byPath(name)
 		return w.addWatch(name, info.dirFlags|unix.NOTE_DELETE|unix.NOTE_RENAME, true)
+	}
+
+	if w.dirsOnly {
+		return filepath.Clean(name), nil
 	}
 
 	// Watch file to mimic Linux inotify.
--- a/backend_kqueue_test.go
+++ b/backend_kqueue_test.go
@@ -8,6 +8,34 @@ import (
 	"testing"
 )
 
+func TestDirsOnly(t *testing.T) {
+	var (
+		tmp = t.TempDir()
+		dir = join(tmp, "dir")
+	)
+	mkdir(t, dir)
+	touch(t, dir, "file1")
+	touch(t, dir, "file2")
+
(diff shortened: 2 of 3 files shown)


arp242:
This will only work for kqueue, so in its current state it's kind of a non-starter on account of that. I'm not sure if it's possible to accurately make this work for all backends.

WhyNotHugo:
`inotify` doesn't currently subscribe to events for each individual file; subscribing to a directory itself suffices. There's really nothing to implement for that backend (I've no idea about others).

I would describe this feature as "it won't deliver Write events on some platforms", keeping clients aware that they may receive superfluous events on some platforms.

arp242:
The entire point of fsnotify is to provide a single interface for different platforms, so you can develop code on Linux and have it work on BSD or Windows, or vice versa, as much as possible.

Having some sort of way to filter events is useful, but this should be a generic solution that works everywhere and not a quick kqueue-specific solution. For example something like:

    fsnotify.WithFilter(fsnotify.Dir)

    fsnotify.WithFilter(fsnotify.File | fsnotify.Symlink)

(just quickly from the top of my head, doesn't need to be exactly this):

On some platforms it can just filter the events. On others (e.g. kqueue) we can optimise and never listen in the first place.

WhyNotHugo:
> ```
> fsnotify.WithFilter(fsnotify.Dir)
> 
> fsnotify.WithFilter(fsnotify.File | fsnotify.Symlink)
> ```

I'm not sure this fits my intended use case: I want events for both files and directories, but I only need events for file creation and deletion (files are never modified inside a Maildir).

I'm not really looking for a kqueue-specific solution: I'm supporting inotify as well. The inotify backend one works fine: fsnotify isn't creating a single fd for every file, and I can ignore superfluous events (which don't happen anyway).

With kqueue, the watcher needs hundreds of thousands of additional file descriptors for those superfluous events.

---

A better fit is something that lets me opt-out of `Write`/`Chmod` events… but the problem is that creating a new file is technically a `Write` event on the directory (and the docs say in some cases that's emitted when a new file is created).

arp242:
If you want very specific things like that, then using kqueue directly is probably best instead of a generic library like fsnotify. That's always going to be fairly limited because it has the limitations of all platforms.

WhyNotHugo:
Understood.

I've probably have to re-implement 90% of fsnotify for this if I want to keep the cross-platform portability (and I do, since I rely on the inotify implementation daily). The only issue with fsnotify is that the kqueue implementation implicitly sets up recursive watchers.

WhyNotHugo:
> The entire point of fsnotify is to provide a single interface for different platforms

FWIW, this is not currently the case: on Linux a single fd is required, whereas in BSD 1+N fds are required for the same use case. Increasing limits to such a high number typically requires reconfiguration by the superuser.

arp242:
> in BSD 1+N fds are required for the same use case

People have been telling that to the BSD people for like 25 years now. 

I think almost every cross-platform application that makes heavy use of filesystem notifications will come to the conclusion that kqueue is not good for this.

WhyNotHugo:
On Mon, 16 Feb 2026, at 16:33, Martin Tournoij wrote:
> *arp242* left a comment (fsnotify/fsnotify#729) <https://github.com/fsnotify/fsnotify/pull/729#issuecomment-3909111399>
>> in BSD 1+N fds are required for the same use case
>> 
> People have been telling that to the BSD people for like 25 years now.
> 

That would be an issue in cases where someone wants to monitors writes for files (and a very annoying one). But that’s not my scenario.

> I think almost every cross-platform application that makes heavy use of filesystem notifications will come to the conclusion that kqueue is not good for this.
> 

The limitation here is in how fsnotify implicitly watches children. Using kqueue directly works fine.

PR #733, opened 2026-04-05
complete recursive watch support

frizbitts:
This PR addressed the open issues regarding the recursive watch feature.
The added tests pass for all supported systems and backends and enableRecurse is set to true.

This still needs some manual testing and a closer look as I am not experienced in golang.

### DIFF
--- a/backend_fen.go
+++ b/backend_fen.go
@@ -12,6 +12,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"time"
 
@@ -26,8 +27,9 @@ type fen struct {
 
 	mu      sync.Mutex
 	port    *unix.EventPort
-	dirs    map[string]Op // Explicitly watched directories
-	watches map[string]Op // Explicitly watched non-directories
+	dirs    map[string]Op        // Explicitly watched directories
+	watches map[string]Op        // Explicitly watched non-directories
+	recurse map[string]struct{}  // Root paths of recursive watches
 }
 
 var defaultBufferSize = 0
@@ -39,6 +41,7 @@ func newBackend(ev chan Event, errs chan error) (backend, error) {
 		Errors:  errs,
 		dirs:    make(map[string]Op),
 		watches: make(map[string]Op),
+		recurse: make(map[string]struct{}),
 	}
 
 	var err error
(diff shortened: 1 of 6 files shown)


mattn:
AFAICS, I’m seeing at least two regressions in the recursive paths:

- kqueue: creating a new subdirectory under a recursive watch appears to install the watches but skip the directory’s own `CREATE` event.
- FEN: for `mkdir -p a/b/c` under a recursive watch, only the first directory seems to emit `CREATE`, while nested directories are silently added.

frizbitts:
Thanks, that should be fixed now.

mattn:
Thanks for the work here, but this needs to be closed and resubmitted as smaller PRs. It bundles four independent changes (inotify prefix fix, FEN recursive, kqueue recursive, default flip) and there are correctness bugs in the kqueue path I won't merge as-is.

**kqueue:**

- `flagRecurse` is set on every subdirectory, not just the root. `Remove("/root/sub")` then routes into `removeRecursive` and tears down the subtree while `w.recurse[root]` stays intact. `isRecurse` should consult `w.recurse`.
- `flagByUser` duplicates the existing `watches.byUser` map. Pick one.
- `removePaths` doesn't delete from `w.path`/`w.wd`, so `w.remove` re-enters via `watchesInDir` → `w.Remove` and walks the same paths repeatedly. Works only because redundant calls fall through to `ErrNonExistentWatch`.

**FEN:** `addRecursive` uses `with.op`, `addRecursiveSubdir` hard-codes `Create|Write|Remove|Rename|Chmod`. A user passing `WithOps(Write)` gets every op on later-created dirs.

**inotify:** the rename handler updates `ww.path` but never re-keys `w.watches.path[oldPath]`. `WatchList()` returns stale paths after a directory rename. Pre-existing, but you're editing this exact loop.

**Tests:** only existing tests enabled on more platforms. The new code paths need targeted tests — `mkdir -p a/b/c` under a recursive root, directory rename inside one, `Remove` on an intermediate path.

**Default flip:** flipping \`enableRecurse\` is user-visible on Linux/Windows. Needs a CHANGELOG entry and a \`Watcher.Add\` doc update.

Suggested split: (1) inotify prefix fix, (2) FEN recursive, (3) kqueue recursive, (4) default flip + docs.

frizbitts:
closed in favor of
https://github.com/fsnotify/fsnotify/pull/752
https://github.com/fsnotify/fsnotify/pull/751
https://github.com/fsnotify/fsnotify/pull/750
https://github.com/fsnotify/fsnotify/pull/753

PR #734, opened 2026-04-13
kqueue: skip broken symlinks in watchDirectoryFiles

### LINKED ISSUE #727: Watching a directory fails when it encounters a symbolic link to a nonexistent file
### Describe the bug

`fsnotify.Watcher.Add` with a directory instead of file (at least in `kqueue` implementation) ignores all files after the first symbolic link that points to a nonexistent file and returns a `file does not exist` error.

[This change](https://github.com/fsnotify/fsnotify/pull/681/changes#diff-cb5d0da6868fb1bd63f13e50565c5ca4951541c055c62707fc016f632e721f2eR413) is suspected.

### Code

```go
package main

import (
)

func check(err error) {
	if err != nil {
		panic(err)
	}
}

func main() {
	check(err)
	defer w.Close()

	err = w.Add("/another/directory")
	check(err)
}
```

### Expected result

Exit code: 0

### Actual result

```none
panic: "/another/directory/aaa": no such file or directory
```

> [!NOTE]
> These steps should be made before running the code.

1. 
    ```sh
    $ touch /some/directory/aaa
    ```
1.
    ```sh
    $ ln -s /another/directory/aaa /some/directory/aaa
    ```
1.
    ```sh
    $ rm /some/directory/aaa
    ```

```none
ProductName:            macOS
ProductVersion:         26.1
BuildVersion:           25B78
```

1.9.0

Yes

Yanhu007:
Fixes #727

## Problem

`watchDirectoryFiles` calls `DirEntry.Info()` which follows symlinks. When a symlink points to a nonexistent target, `Info()` returns an error that propagates up, aborting the entire directory watch. All files listed after the broken symlink are never watched.

## Fix

When `Info()` fails and the entry is a symlink, skip it and continue processing remaining files. This is consistent with the existing pattern for permission errors (EACCES/EPERM) which also skip the file rather than aborting.

```go
fi, err := f.Info()
if err != nil {
    if f.Type()&os.ModeSymlink != 0 {
        continue  // skip broken symlinks
    }
    return fmt.Errorf("%q: %w", path, err)
}
```

### DIFF
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -573,6 +573,12 @@ func (w *kqueue) watchDirectoryFiles(dirPath string) error {
 
 		fi, err := f.Info()
 		if err != nil {
+			// Broken symlinks (pointing to nonexistent targets) will
+			// fail here. Skip them instead of aborting the entire
+			// directory watch, so remaining files are still watched.
+			if f.Type()&os.ModeSymlink != 0 {
+				continue
+			}
 			return fmt.Errorf("%q: %w", path, err)
 		}
 


mattn:
I don't think this fixes it. `DirEntry.Info()` uses `lstat` on Unix, so it succeeds for broken symlinks — the added `continue` isn't reached. I verified locally: `f.Info()` returns no error on a dangling symlink, while `os.Open` fails with ENOENT.

The real failure is `unix.Open` at backend_kqueue.go:381 reached via `addWatch(..., listDir=true)`, which surfaces from line 589 (same `%q: %w` format as the `f.Info()` branch, likely the source of confusion). Also, no test is added, so this shouldn't be merged as-is.

PR #736, opened 2026-04-21
windows: fix nil pointer dereference in remWatch

mattn:
Fix nil pointer dereference in `remWatch` when called on a non-watched path with the `\...` recurse suffix. The nil check ran after `watch.recurse` was dereferenced; reorder the check and close the handle in the early-return paths. Includes a regression test.


shogo82148 (approved):
LGTM

mattn:
macos-13 is already closing down. So merge without tests for macros-13

https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

PR #737, opened 2026-04-21
bump macos-14

shogo82148:
macos-13 is no longer available.

- https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/

so, we need to use `macos-14`.


mattn (approved):

mattn:
Thank you but closing by https://github.com/fsnotify/fsnotify/pull/738

PR #740, opened 2026-04-22
kqueue: drop watches directly in Close() instead of going through remove()

### LINKED ISSUE #732: kqueue: Close() leaks all watch file descriptors
Close() in backend_kqueue.go marks the watcher as closed via shared.close() before iterating the remove loop. Then remove() checks isClosed() at entry and returns nil without closing any file descriptors.

On macOS with kqueue, watching a directory also opens individual file descriptors for every file in the directory. A watcher monitoring N directories with M total files leaks N+M descriptors on every Close() call. Long-running processes that recreate watchers (e.g. dev servers with hot reload) accumulate leaked descriptors until hitting EMFILE ("too many open files").

Discovered by inspecting a Go dev server that hit 10,376 open fds after ~9 hot reload cycles, each leaking ~1,400 file/directory descriptors from watcher.Close().

Version: v1.9.0

Repro:
1. Create a watcher with Add() on a directory containing many files
2. Close() the watcher
3. Check /proc/self/fd or lsof: all watched file/directory fds are still open

The fix is to close watch descriptors directly in Close() by iterating the wd map instead of delegating to Remove() which short-circuits on isClosed():

https://github.com/aperturerobotics/fsnotify/pull/1

SAY-5:
## Summary

Fixes #732.

`kqueue.Close()` calls `w.Remove(name)` in a loop *after* `w.shared.close()` has marked the watcher closed. `w.Remove` routes through `w.remove`, which short-circuits on `w.isClosed()` and returns nil without calling `unix.Close` on the watch fd or dropping the watches bookkeeping. Every directory watch (and each file inside it — a dir watch opens one fd per child on macOS) leaked its descriptor on `Close`, so long-running processes that recycle watchers (hot-reload dev servers, test harnesses) hit `EMFILE` after a few cycles.

## Fix

Iterate the paths directly in `Close()`: register the `EV_DELETE` kevent, `unix.Close` the fd, and drop the `watches` entry. `kqueue.remove()` is unchanged, so `Remove()` on a live watcher keeps the same semantics; only `Close()` stops leaning on it.

## Tests

Adds `TestCloseClearsWatchState` in `backend_kqueue_test.go`: creates a watcher over a temp dir + file, calls `Close`, and asserts every `watches` map is drained.

- **On master:** fails — `watches.path: 3 entries`, `watches.wd: 3 entries`, `watches.byUser: 2 entries`, `watches.byDir: 3 entries`.
- **With the fix:** passes.

`go test -count=1 ./...` green on darwin; `go vet` clean.

### DIFF
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -246,9 +246,22 @@ func (w *kqueue) Close() error {
 		return nil
 	}
 
+	// Snapshot and drop all watches directly. w.Remove -> w.remove
+	// short-circuits on isClosed() (which is already true after
+	// w.shared.close() above), so calling Remove here in the happy path
+	// leaked every watched directory + file descriptor. On macOS a
+	// single directory watch opens an fd for every file in the dir, so
+	// long-running processes that recreate watchers (hot-reload dev
+	// servers, etc.) ran out of fds with EMFILE (#732).
 	pathsToRemove := w.watches.listPaths(false)
 	for _, name := range pathsToRemove {
-		w.Remove(name)
+		info, ok := w.watches.byPath(name)
+		if !ok {
+			continue
+		}
+		_ = w.register([]int{info.wd}, unix.EV_DELETE, 0)
+		unix.Close(info.wd)
+		w.watches.remove(info.wd, name)
 	}
 
 	unix.Close(w.closepipe[1]) // Send "quit" message to readEvents
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -257,6 +257,10 @@ func (w *kqueue) Close() error {
 	for _, name := range pathsToRemove {
 		info, ok := w.watches.byPath(name)
 		if !ok {
+			// w.path has an entry for name but w.wd doesn't --
+			// drop the stale lookup entry so the map state is
+			// consistent after Close.
+			w.watches.remove(0, name)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


backend_kqueue.go
```diff
@@ -246,9 +246,22 @@ func (w *kqueue) Close() error {
...
+	// w.shared.close() above), so calling Remove here in the happy path
+	// leaked every watched directory + file descriptor. On macOS a
+	// single directory watch opens an fd for every file in the dir, so
+	// long-running processes that recreate watchers (hot-reload dev
+	// servers, etc.) ran out of fds with EMFILE (#732).
 	pathsToRemove := w.watches.listPaths(false)
 	for _, name := range pathsToRemove {
-		w.Remove(name)
+		info, ok := w.watches.byPath(name)
+		if !ok {
```
mattn, line 259:
```
		if !ok {
			w.watches.remove(0, name)
			continue
		}
```

mattn (changes requested):

SAY-5:
Thanks @mattn — applied your suggestion. The not-ok branch now calls `w.watches.remove(0, name)` so stale `w.path` / `w.byUser` entries get purged even when `byPath` can't resolve to a live `w.wd` entry. `go test -count=1 ./...` green locally.

mattn (approved):
I'll look this in later again. Thank you

mattn:
Thank you

PR #741, opened 2026-04-22
inotify: avoid copying event buffers when reading names

mattn:
This avoids copying the full inotify read buffer when extracting event names. Instead of copying the 64KB buffer and trimming trailing NUL bytes with `strings.TrimRight`, it now scans the name directly from the original buffer and trims the trailing NULs in place.

Before: `1151 ns/op`
After: `31 ns/op`


shogo82148 (approved):
LGTM

PR #742, opened 2026-04-22
windows: restore TestRemoveState

mattn:
TestRemoveState was skipped because it counted entries in the Windows backend's top-level `watches` map, which does not match the nested structure the backend actually uses. Count watched paths via `WatchList()` instead so the assertion reflects the real number of watch entries.

Verified with `go test -run TestRemoveState -count=20 -v` on Windows.


shogo82148 (approved):
ﾖｼｯ

PR #743, opened 2026-04-22
windows: re-enable chanClosed check in TestClose/events_not_read

mattn:
The `chanClosed` assertion in `TestClose/events_not_read` was skipped on Windows, but the check now passes reliably. Drop the `runtime.GOOS` guard so the behaviour is exercised on Windows as well.

Verified with `go test -run 'TestClose/events_not_read' -count=20 -v` on Windows.


shogo82148 (approved):
LGTM

PR #744, opened 2026-04-22
all: bump minimum Go version to 1.23

mattn:
Bumps the `go` directive in `go.mod` from 1.19 to 1.23 and the CI matrix likewise. `reflect.SliceHeader` has been deprecated since Go 1.21, so the two remaining uses in `backend_windows.go` are rewritten using `unsafe.Slice` / `unsafe.SliceData` (one of them already carried a `TODO: Use unsafe.Slice` comment). This also silences the staticcheck SA1019 failures that surface as soon as the go directive is bumped, which came up while looking at #723.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-22.04', 'ubuntu-latest', 'ubuntu-24.04-arm']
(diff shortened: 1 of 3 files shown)


mattn:
Will update CHANGELOG.md (for the minimum Go version).

shogo82148 (approved):
LGTM

PR #745, opened 2026-04-22
windows: document directory Write events and stabilize tests

mattn:
Document the Windows recursive-watch behaviour where creating, renaming, or removing a child entry on an NTFS-backed volume can surface as a `Write` on the containing directory (kqueue has the same "directory `Write` = contents changed" semantics; inotify does not).

Add a `cmpEvents` normalization that drops such directory `Write`s when a descendant event is in the same batch, so tests like `testdata/watch-recurse/rename-dir` are no longer flaky.

The normalization is interim; a follow-up will replace it with an "optional event" syntax in the testdata script format. Supersedes #739.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,14 +172,12 @@ distro's documentation):
     fs.inotify.max_user_instances=256
 
 ### Windows
-The Windows backend uses `ReadDirectoryChangesW`, which reports changes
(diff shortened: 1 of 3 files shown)


shogo82148 (approved):
LGTM

PR #748, opened 2026-04-26
kqueue: skip ENOENT entries in watchDirectoryFiles

### LINKED ISSUE #727: Watching a directory fails when it encounters a symbolic link to a nonexistent file

mattn:
Fixes #727. On kqueue, `Watcher.Add` on a directory containing a dangling symlink failed because `watchDirectoryFiles` only swallowed `EACCES`/`EPERM` from `internalWatch`. When `unix.Open(O_EVTONLY)` followed a symlink to a missing target it returned `ENOENT`, which propagated up and aborted the whole `Add`. Add `os.ErrNotExist` to the same skip path so the entry is marked as seen and watching continues for the rest of the directory. Regression from #681. Adds a regression test under `testdata/watch-dir/bug-727`.

### DIFF
--- a/backend_kqueue.go
+++ b/backend_kqueue.go
@@ -578,12 +578,14 @@ func (w *kqueue) watchDirectoryFiles(dirPath string) error {
 
 		cleanPath, err := w.internalWatch(path, fi)
 		if err != nil {
-			// No permission to read the file; that's not a problem: just skip.
-			// But do add it to w.fileExists to prevent it from being picked up
-			// as a "new" file later (it still shows up in the directory
+			// No permission, or the entry resolved to a missing target
+			// (e.g. a dangling symlink): not a problem, just skip. But
+			// do mark it as seen to prevent it from being picked up as
+			// a "new" file later (it still shows up in the directory
 			// listing).
 			switch {
-			case errors.Is(err, unix.EACCES) || errors.Is(err, unix.EPERM):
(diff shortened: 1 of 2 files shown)


shogo82148 (approved):
ﾖｼｯ

PR #749, opened 2026-04-26
windows: lock watch field updates against concurrent WatchList

### LINKED ISSUE #709: Race between Add() and WatchList() on Windows since v1.9.0
### Describe the bug

The Go race detector flags on concurrent read/write across goroutines when one is calling .Add() while the other calls .WatchList(). The issue was not present on v1.7.0 using the same test, but was introduced with v1.9.0 and persists on the main branch at v1.9.1-0.20250811202840-88b7b41337a8. I was unable to test on v1.8.0 as the same test just hangs.

### Code to Reproduce

```
go get github.com/fsnotify/fsnotify@main
```

```go
//go:build windows
// +build windows

package fsnotify_race_test

import (
	"os"
	"path/filepath"
	"testing"

)

// Reproduces: readDirChangesW.addWatch (write) vs WatchList (read).
func TestWatchListRace(t *testing.T) {
	root := t.TempDir()

	if err != nil {
		t.Fatalf("NewWatcher: %v", err)
	}
	defer w.Close()

	if err := w.Add(root); err != nil {
		t.Fatalf("Add(root): %v", err)
	}

	// --- reader: continuously calls WatchList()
	done := make(chan struct{})
	go func() {
		for {
			select {
			case <-done:
				return
			default:
				_ = w.WatchList()
			}
		}
	}()

	// --- writer: Add() call => addWatch()
	dir := filepath.Join(root, "test")
	os.Mkdir(dir, 0o755)
	_ = w.Add(dir)

	close(done)
}
```

N/A

OS Name:                       Microsoft Windows 11 Pro
OS Version:                    10.0.26100 N/A Build 26100
OS Manufacturer:               Microsoft Corporation
OS Configuration:              Standalone Workstation
OS Build Type:                 Multiprocessor Free

v1.9.0

Yes

mattn:
Fixes #709. Calling `WatchList()` concurrently with `Add()`/`Remove()` on Windows raced with the I/O thread mutating `watchEntry.mask` and `watchEntry.names` outside the lock; the race detector caught the `mask` access but the more dangerous case is concurrent map iteration vs. write on `watchEntry.names`, which can panic at runtime. Hold `w.mu` around all writes to those fields (`addWatch`, `remWatch`, `deleteWatch`, and the rename/remove paths in `readEvents`). `deleteWatch` snapshots the map and mask under the lock and then calls `sendEvent` outside it to avoid blocking on the Events channel while holding the mutex. Adds `TestWatchListRace` as a regression test.

### DIFF
--- a/backend_windows.go
+++ b/backend_windows.go
@@ -358,22 +358,26 @@ func (w *readDirChangesW) addWatch(pathname string, flags uint64, bufsize int) e
 	} else {
 		windows.CloseHandle(ino.handle)
 	}
+	w.mu.Lock()
 	if pathname == dir {
 		watchEntry.mask |= flags
 	} else {
 		watchEntry.names[filepath.Base(pathname)] |= flags
 	}
+	w.mu.Unlock()
 
 	err = w.startRead(watchEntry)
 	if err != nil {
 		return err
 	}
 
+	w.mu.Lock()
 	if pathname == dir {
 		watchEntry.mask &= ^provisional
 	} else {
 		watchEntry.names[filepath.Base(pathname)] &= ^provisional
 	}
+	w.mu.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


shogo82148 (approved):
ﾖｼｯ

PR #750, opened 2026-04-29
inotify fixes

frizbitts:
split patch from #733

### DIFF
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -102,8 +102,9 @@ func (w *watches) removePath(path string) ([]uint32, error) {
 
 	wds := make([]uint32, 0, 8)
 	wds = append(wds, wd)
+	prefix := path + "/"
 	for p, rwd := range w.path {
-		if strings.HasPrefix(p, path) {
+		if strings.HasPrefix(p, prefix) {
 			delete(w.path, p)
 			delete(w.wd, rwd)
 			wds = append(wds, rwd)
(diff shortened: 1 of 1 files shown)


arp242:
It's not clear to me what exactly this fixes, but I think this may fix the same?  https://github.com/fsnotify/fsnotify/commit/853adeb28b4e7b01cf719f12619fcdeea8c06d48#diff-700514c760b074d75ca5b007023944850aeaf3eaf4f116f21e42a7e1cc32b05cR504-R516

If this fixes something else/more, then having a description and test case would be good.

frizbitts:
yes its the same issue
Closing this as it is fixed in 853adeb

@arp242 Please check #751 #752 #753 as well and let me know if they needs any changes.
Im looking forward to a release with recursive watch support sometime soon.

arp242:
Okay, thanks. I didn't see this PR before I made those changes – it was just something I encountered and fixed.

The kqueue PR is a bit awkward since it duplicates my kqueue rewrite that I did last year (kq branch). That also fixes some other issues like event filtering, rename tracking, as well as some of the other reported bugs. I stopped working on that because a FreeBSD update broke Go and no longer had a working environment, so I never really resumed my work on that. So not sure what the best way forward on that is (merge your PR or continue that).

For FEN ... I'm not sure either. I've been somewhat thinking about dropping FEN support altogether since illumos supports inotify, so I think maybe we can just use that instead of adding another backend? Need to see how well it actually works though.

frizbitts:
>  So not sure what the best way forward on that is (merge your PR or continue that).

I don't mind either way. Same for FEN. The reason I opened those PRs was that previous comments hinted on full support as a condition for enabling recursive watches in the release builds https://github.com/fsnotify/fsnotify/pull/753
In the end I only need the Linux backend myself.
let me know when you know how to move forward in that direction and I will work on that.

PR #754, opened 2026-05-01
inotify: don't remove sibling watches sharing a path prefix

mattn:
`watches.removePath` used `strings.HasPrefix(p, path)` to find descendants of a recursive watch, so removing `/tmp/a/...` also dropped a sibling watch on `/tmp/ab`. Match on a path-separator boundary so only the target and its real descendants are removed.


mattn:
Plan to cut v1.10.1 once this lands.

mattn:
Spotted the same prefix bug in two more places: recursive rename in `backend_inotify.go:511` and `backend_windows.go:608`. Will send a separate PR.

shogo82148 (approved):
ﾖｼｯ

mattn:
Thank you

PR #755, opened 2026-05-01
inotify, windows: don't rename sibling watches sharing a path prefix

mattn:
The recursive rename path-rewriting loops used `strings.HasPrefix(ww.path, old)` to decide which watches belong under a renamed directory, so renaming `/tmp/a` to `/tmp/x` also rewrote a sibling watch on `/tmp/ab` to `/tmp/xb` (and on Windows, `C:\root\a` → `C:\root\x` rewrote `C:\root\ab` to `C:\root\xb`). Subsequent events were then misrouted. Match on a path-separator boundary so only the renamed directory and its real descendants are rewritten.

Stacked on top of #754 (which fixes the same shape of bug in `removePath`); merging that one first will simplify the diff.

### DIFF
--- a/backend_inotify.go
+++ b/backend_inotify.go
@@ -508,7 +508,7 @@ func (w *inotify) handleEvent(inEvent *unix.InotifyEvent, buf *[65536]byte, offs
 					if k == watch.wd || ww.path == ev.Name {
 						continue
 					}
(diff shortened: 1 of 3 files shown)


shogo82148 (approved):
ﾖｼｯ

PR #756, opened 2026-05-06
inotify: fix WatchList after rename, clean tests

arp242:
WatchList was incorrect after renaming: instead of only updating watches.wd loop over watches.path instead of watches.wd and update both.

Also properly do some tests that were added: fsnotify behaviour should be cross-platform, so tests should run everywhere. Backout some unclear and factually incorrect (seemingly AI-generated) "documentation", and TestRemoveState() on Windows wasn't actually fixed (changing what a test tests doesn't fix the problem).


Inndy:
You forget to mention "Remove mattn as maintainer, hijack whole organization" in PR and commit message, should be documented since it's a breaking change

arp242:
There seems to be some misinformation making the rounds about all of this; the discussion for that is over here: https://github.com/fsnotify/fsnotify/issues/757#issuecomment-4399405186

Locking as having a discussion in two places is not helpful.

PR #766, opened 2026-07-23
fix: nil Watcher doesn't panic on Add/Remove/Close

sonnemusk:
```go
var w *fsnotify.Watcher
w.Add("/tmp") // panic
```

Return an error (or nil for Close/WatchList) instead.


sonnemusk:
Closing this — it was a defensive nil guard, not something I hit in real use. Callers should not pass nil here; panicking (or failing at the call site) is fine.

PR #769, opened 2026-07-28
windows: don't close a torn-down watch's handle twice

### LINKED ISSUE #768: windows: double CloseHandle on a torn-down watch aborts the Go runtime (ERROR_INVALID_HANDLE)
### Describe the bug

On Windows, `readEvents` can close the same directory handle twice. Because Windows reissues a closed handle's numeric value to the next caller — and in a Go program the most frequent caller is the runtime itself — the second close destroys a handle the *runtime* owns, and the process later dies with a fatal error whose stack points nowhere near fsnotify:

```
runtime: waitforsingleobject wait_failed; errno=6
fatal error: runtime.semasleep wait_failed
```

`errno=6` is `ERROR_INVALID_HANDLE`. I have also seen `runtime: setevent failed; errno=6` / `fatal error: runtime.semawakeup`, and `runtime: netpoll failed` — three different runtime primitives, one cause: the M-parking event handles the runtime allocates in `semacreate` (`CreateEvent`), and the netpoll port, are exactly the objects that land on a handle value fsnotify just freed.

This may explain part of the existing unexplained Windows cluster — #704, #669, #502 — though I have only proven the mechanism below, not that it is behind each of those.

### The mechanism

`readEvents` re-arms the watch unconditionally at the end of the event-processing loop:

```go
if err := w.startRead(watch); err != nil {
	w.sendError(err)
}
```

but `watch` is recovered from the completion packet's `OVERLAPPED` pointer:

```go
watch := (*watch)(unsafe.Pointer(ov))
```

and that packet may have been queued *before* a `remWatch` tore the watch down. `remWatch` sets `watch.mask = 0` and calls `startRead`, which takes the `mask == 0` branch — closing `watch.ino.handle` and deleting the watch from `w.watches`. A read that had already completed with data is still sitting in the completion queue (`CancelIo` does not retract packets that already completed). When it is dequeued, the tail re-arm calls `startRead` on that dead watch: `CancelIo` fails on the closed handle, `deleteWatch` zeroes the mask, and `CloseHandle(watch.ino.handle)` runs a **second** time.

### There is a second, causally linked bug

`sendError` receives `Close`'s handshake token from `w.done` and drops it, where `sendEvent` puts it back:

```go
select {
case <-w.done:      // consumed, never returned
	return false
```

`w.done` has capacity 1 and carries the reply channel that is `Close`'s only handshake with the reader. If an error is reported while a `Close` is in flight, the token is lost: the reader falls through to `default`, parks in `GetQueuedCompletionStatus` having already consumed the wakeup, and `Close` blocks forever on `<-ch`. Its reader thread, watches and handles leak.

The two are linked: the double close is what makes `CancelIo` fail, which is what calls `sendError`, which is what loses the token. So the first bug is a frequent *cause* of the second.

### Evidence

Reproduced on a Windows **arm64** host (Go 1.26.5), driving repeated `Add`/`Remove` of one directory that a second goroutine writes to every 1 ms — the shape of a watcher that subscribes and unsubscribes under live file activity.

Against a build that routes every `CloseHandle` in `backend_windows.go` through a tracker, a reported double close carries the **same watcher and the same `*watch`** in both stacks:

```
=== DOUBLE CLOSE of handle 0x1a8 (CloseHandle err=The handle is invalid.) ===
--- first close was here ---
fsnotify.trackedClose(0x1a8)
fsnotify.(*readDirChangesW).startRead(0x62b2f7814100, 0x62b2f7926500)   backend_windows.go:450
fsnotify.(*readDirChangesW).remWatch(0x62b2f7814100, ...)               backend_windows.go:419
fsnotify.(*readDirChangesW).readEvents(0x62b2f7814100)                  backend_windows.go:524
--- second close is here ---
fsnotify.trackedClose(0x1a8)
fsnotify.(*readDirChangesW).startRead(0x62b2f7814100, 0x62b2f7926500)   backend_windows.go:450
fsnotify.(*readDirChangesW).readEvents(0x62b2f7814100)                  backend_windows.go:642
```

Measured over 150 rounds of that churn:

| | before | after the patch |
|---|---|---|
| double closes | **2404** | **0** |
| hung `Close`s per 200 watcher rounds | **4** | **0** |
| process handle count | slow upward drift | flat |

And in the application that found this (a GUI toolkit whose file-watching layer does exactly this Add/Remove churn), the Go runtime fatal reproduced within **3** test iterations before the patch and **not once in 160** runs after.

Note that the handle count barely moves — this is not handle exhaustion. `ERROR_INVALID_HANDLE` on a handle the runtime already owns is a stray close.

### Versions

- fsnotify v1.9.0, and the code is unchanged on `main` at v1.10.1 — both bugs are still present there. Verified by reading `backend_windows.go` on `main`; the reproduction ran against v1.9.0.
- Go 1.26.5, windows/arm64. Nothing about the mechanism is arm64-specific; that is just where the timing made it frequent.

### Fix

PR follows: drop completion packets whose watch is no longer registered, and put the `w.done` token back in `sendError`.

trendvidia:
Fixes #768, where the diagnosis and the measurements live.

Two changes to `backend_windows.go`.

**1. Drop completions for a watch that is no longer registered.**

`readEvents` re-arms the watch unconditionally at the end of the event loop, but `watch` is recovered from the completion packet's `OVERLAPPED` pointer and that packet may predate a `remWatch` that already closed its handle and dropped it from `w.watches`. The re-arm then runs `startRead` on a dead watch: `CancelIo` fails on the closed handle, `deleteWatch` zeroes the mask, and `CloseHandle` runs a second time on the same handle.

Windows reissues a closed handle's numeric value to the next caller. In a Go program the most frequent caller is the runtime itself — `CreateEvent` per new M in `semacreate` — so the stray close destroys a handle the runtime believes it owns, and the next `WaitForSingleObject`/`SetEvent`/netpoll on it aborts the whole process with `ERROR_INVALID_HANDLE`, from a stack pointing nowhere near fsnotify.

The check goes before the `switch qErr` so it covers the `ERROR_ACCESS_DENIED` branch too, which also calls `deleteWatch` + `startRead`. Identity comparison (`== watch`) rather than a presence test, so a *new* watch for the same directory — same volume and file index — does not make a stale packet look live.

**2. Put `Close`'s handshake token back in `sendError`.**

`w.done` has capacity 1 and carries the reply channel that is `Close`'s only handshake with the reader. `sendEvent` receives it and puts it back; `sendError` received it and dropped it. An error reported while a `Close` is in flight therefore lost the token: the reader fell through to `default`, parked in `GetQueuedCompletionStatus` having already consumed the wakeup, and `Close` blocked forever on a reply that could never come — leaking the reader thread and its handles.

The two are linked: the double close is what makes `CancelIo` fail, which is what calls `sendError`, which is what loses the token. Fixing only the first would make the second much rarer without removing it.

**Measured** on windows/arm64 (Go 1.26.5), 150 rounds of repeated `Add`/`Remove` of one directory that another goroutine writes to every 1 ms, against a build that routes every `CloseHandle` through a tracker:

| | before | after |
|---|---|---|
| double closes | 2404 | **0** |
| hung `Close`s per 200 watcher rounds | 4 | **0** |
| process handle count | slow upward drift | flat |

In the downstream application that surfaced this, the Go runtime fatal reproduced within 3 test iterations before and did not appear in 160 runs after.

Builds for windows, linux and darwin; `gofmt` clean. I could not run the Windows test suite on this arm64 box beyond the targeted reproduction, so a CI run on windows/amd64 is worth having before merge.

### DIFF
--- a/backend_windows.go
+++ b/backend_windows.go
@@ -80,7 +80,12 @@ func (w *readDirChangesW) sendError(err error) bool {
 		return true
 	}
 	select {
-	case <-w.done:
+	case ch := <-w.done:
+		// Put the token back, the way sendEvent does. It is Close's only handshake
+		// with the reader: swallowing it here leaves Close blocked forever on a
+		// reply that can never come, while the reader parks in
+		// GetQueuedCompletionStatus having already consumed the wakeup.
+		w.done <- ch
 		return false
 	case w.Errors <- err:
 		return true
@@ -540,6 +545,22 @@ func (w *readDirChangesW) readEvents() {
 				}
 			default:
 			}
+			continue
+		}
+
+		// Drop completions for a watch that is no longer registered. remWatch may
+		// have torn this watch down between the read completing and us dequeuing
+		// it: its handle is closed and it is out of w.watches, but the completion
+		// packet queued earlier still carries its address. Falling through would
+		// re-arm it below and close that handle a SECOND time -- and Windows may
+		// have already reissued the value. In a Go program the most frequent taker
+		// is the runtime itself (CreateEvent per new M in semacreate), so the stray
+		// close destroys a handle the runtime believes it owns and the next
+		// semasleep/semawakeup/netpoll on it dies with ERROR_INVALID_HANDLE.
+		w.mu.Lock()
+		live := w.watches.get(watch.ino) == watch
+		w.mu.Unlock()
+		if !live {
 			continue
 		}
 


MikeeI:
The new stale-completion guard appears to dereference a `watch` after the code has removed the last Go-owned reference that keeps that allocation alive.

`startRead` calls `CancelIo`, closes the handle, and deletes the watch from `w.watches` immediately when the mask reaches zero. [`CancelIo`](https://learn.microsoft.com/en-us/windows/win32/fileio/cancelio) only requests cancellation; the canceled operation still completes through the normal completion mechanism. The queued IOCP packet retains the numeric `OVERLAPPED` address, but that raw kernel pointer is not a Go GC root.

When the cancellation packet is later dequeued, this PR converts `ov` back to `*watch` and evaluates `watch.ino` in the new registry check before reaching the existing `ERROR_OPERATION_ABORTED` branch. If GC and allocator reuse occur between map deletion and dequeue, that dereference can read reclaimed/reused storage, crash, or mistake a replacement allocated at the same address for the live watch. Before this change, an `ERROR_OPERATION_ABORTED` packet could reach the switch and `continue` without dereferencing fields from the recovered pointer.

The stale-packet fix therefore needs to keep retired watches and their `OVERLAPPED` storage reachable in a Go-owned pending registry until the corresponding completion has been dequeued. The same lifecycle also needs to prevent reusing an `OVERLAPPED` while the canceled operation's completion is still outstanding.

The `sendError` token restoration looks consistent with `sendEvent`. I also confirmed the branch cross-compiles for `windows/amd64`; that proves types only, not IOCP ordering or lifetime safety. A useful native stress check would combine `GOGC=1`, forced GC, Add/Remove/re-add churn, a concurrent writer, Close deadlines, and handle/`OVERLAPPED` tracking. On `windows/amd64`, the scoped run should also use `-race`; Windows/arm64 requires the same native stress without the race detector.

### Disclosure

Investigated thoroughly with openai-codex/gpt-5.6-sol (medium reasoning effort), using [Oh My Pi](https://github.com/can1357/oh-my-pi) as the agent framework.

This report is not generic or unreviewed AI-generated output. Its claims were checked against the cited evidence, and it includes the relevant detail intended to help maintainers resolve the issue.

If reports like this are not useful to the project, please let me know and I will refrain from submitting similar ones. My intent is to help without wasting maintainer time or energy or discouraging their work.

Thank you for your work.

PR #770, opened 2026-08-01
Expose close-write watch support

kwkeefer:
## Summary

- export the existing close-write operation as `UnportableCloseWrite`
- export `WithOps` and `Watcher.Supports` so callers can request it safely
- verify with a public-package Linux filesystem test that close-write is emitted only after the child writer closes
- verify the capability query matches supported backends

This is a narrow export of behavior already implemented by the inotify backend; descriptor management and event handling are unchanged.

The patch is based on v1.10.1 because a downstream consumer is pinning that exact release plus this change.


kwkeefer:
Closing and removing the downstream fork; we are taking a different implementation direction.

PR #773, opened 2026-08-28
windows: fix deadlock between Close() and Add()/Remove() (#704)

### LINKED ISSUE #704: windows: Dead lock between watcher.Close() and watcher.Add()
### Describe the bug

There is a race condition between watcher.Close() and watcher.Add() which my leads to deadlock in windows. 

### Code to Reproduce

```
func TestFoo(t *testing.T) {
	root := t.TempDir()

	if err != nil {
		t.Fatal(err)
	}
	if err := watcher.Add(root); err != nil {
		t.Fatal(err)
	}

	done := make(chan struct{})

	go func() {
		<-done
		watcher.Close()
	}()

	for i := range 100 {
		if err := os.Mkdir(filepath.Join(root, fmt.Sprintf("foo%v", i)), 0755); err != nil {
			t.Fatal(err)
		}
		if i == 10 {
			close(done)
		}
		if err := watcher.Add(filepath.Join(root, fmt.Sprintf("foo%v", i))); err != nil && !errors.Is(err, fsnotify.ErrClosed) {
			t.Fatal(err)
		}
	}

}
```

It is not promised to fail every time. You can run it 100 times and normally there is a failure.

I made some local changes to fsnotify and run this test. Here is the code edit I made. Only 4 line of print. The first is when Close() is called, before the readEvents function returns. 

The second is when a registration request arrived. The third is when a registration request is done(when the w.addWatch returns).

Another place is 

Image

Image

Here is the print statement of the test. 

```
PS > go test -run TestFoo -v -count 10
=== RUN   TestFoo
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo0
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo0
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo0
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo1
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo1
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo1
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo2
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo2
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo2
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo3
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo3
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo3
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo4
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo4
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo4
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo5
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo5
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo5
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo6
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo6
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo6
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo7
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo7
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo7
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo8
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo8
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo8
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo9
readEvents capture input: PATH\TO\Temp\TestFoo700788738\001\foo9
readEvents addWatch returned PATH\TO\Temp\TestFoo700788738\001\foo9
AddWith sending in to w.input PATH\TO\Temp\TestFoo700788738\001\foo10
readEvents capture termination signal, return
<---- stuck here
```

The race is happening when both Close() and Add() is being called, it happens, 
1st Add go routine pass the isClosed check, 
2nd close go routine Close the watcher,
3rd close go routine terminate the readEvents so no more receiver of `w.input`
4th Add go routine trying to send the watch request to `w.input`

But because the receiver is terminated, so the sender get stuck forever. We are planning to work this out by making sure wait for all the Add go routine to finish and then call Close() so there is no overlapping between `Add` and `Close`

windows (can not share detail)

1.9.0

No

CAOShurong:
## What

Fixes a deadlock on Windows between a concurrent `watcher.Close()` and `watcher.Add()` (also `Remove()`) — reported in #704.

## Root cause

The reader goroutine in `backend_windows.go` selects on both `w.done` (from `Close()`) and `w.input` (from `Add`/`Remove`). When `Close()`'s `done` signal arrives while an `Add` input is still queued, the goroutine takes the `done` branch and exits — orphaning that pending input. The caller's `AddWith`/`Remove` then blocks forever on `<-in.reply`, hanging the whole test/binary.

## Fix

Add a `closeCh` channel that `Close()` closes (under the existing `mu` lock). `AddWith` and `Remove` now `select` on `closeCh` and return `ErrClosed` instead of blocking when the watcher is shutting down. Because `close()` unblocks every waiter simultaneously, the signal is never lost and there is no double-close hazard.

## Verification

- Reproduced the hang on Windows (Go `net/http` `TestFoo` pattern from #704) before the fix: `AddWith` blocked on `backend_windows.go:138` (`<-in.reply`) until the test binary timed out.
- After the fix, the new regression test `TestWindowsCloseAddRace` runs the race 200 times and passes (~5.5s) on Windows; `Add`/`Remove` return `ErrClosed` promptly during the race instead of blocking.
- `go build ./...`, `go vet ./...` clean.
- Remaining `TestScript/watch-recurse` ordering diffs are **pre-existing** on pristine `main` on this host (verified by stashing the change) and are unrelated to this fix.

## Note on assistance

This change was prepared with AI assistance (Codex); I reproduced, diagnosed, and verified the bug and the fix locally on Windows. Happy to adjust the approach if maintainers prefer a different teardown scheme.

Closes #704

### DIFF
--- a/backend_windows.go
+++ b/backend_windows.go
@@ -32,6 +32,7 @@ type readDirChangesW struct {
 	mu      sync.Mutex // Protects access to watches, closed
 	watches watchMap   // Map of watches (key: i-number)
 	closed  bool       // Set to true when Close() is first called
+	closeCh chan struct{} // Closed (once) when Close() runs, so Add/Remove can bail out
 }
 
 var defaultBufferSize = 50
@@ -48,6 +49,7 @@ func newBackend(ev chan Event, errs chan error) (backend, error) {
 		watches: make(watchMap),
 		input:   make(chan *input, 1),
 		done:    make(chan chan<- error, 1),
+		closeCh: make(chan struct{}),
 	}
 	go w.readEvents()
 	return w, nil
@@ -94,6 +96,7 @@ func (w *readDirChangesW) Close() error {
 
 	w.mu.Lock()
 	w.closed = true
+	close(w.closeCh) // Notify any in-flight Add/Remove that we're shutting down.
 	w.mu.Unlock()
 
 	// Send "done" message to the reader goroutine
@@ -131,11 +134,24 @@ func (w *readDirChangesW) AddWith(name string, opts ...addOpt) error {
 		reply:   make(chan error),
 		bufsize: with.bufsize,
 	}
+	// Bail out if Close() is racing with this Add(): otherwise the reader
+	// goroutine may take the "done" branch and exit, orphaning this input
+	// and leaving Add() blocked forever (see issue #704).
+	select {
+	case <-w.closeCh:
+		return ErrClosed
+	default:
+	}
 	w.input <- in
 	if err := w.wakeupReader(); err != nil {
 		return err
 	}
-	return <-in.reply
+	select {
+	case err := <-in.reply:
+		return err
+	case <-w.closeCh:
+		return ErrClosed
+	}
 }
 
 func (w *readDirChangesW) Remove(name string) error {
@@ -156,7 +172,12 @@ func (w *readDirChangesW) Remove(name string) error {
 	if err := w.wakeupReader(); err != nil {
 		return err
 	}
-	return <-in.reply
+	select {
+	case err := <-in.reply:
+		return err
+	case <-w.closeCh:
+		return ErrClosed
+	}
 }
 
 func (w *readDirChangesW) WatchList() []string {
--- a/backend_windows_test.go
+++ b/backend_windows_test.go
@@ -7,9 +7,13 @@
 package fsnotify
 
 import (
+	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestRemoveState(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Voyagerroc-Lab (changes requested):
In backend_windows.go, the closeCh checks happen after an unconditional send to w.input in AddWith, and Remove has no pre-send close check. Close can make the reader exit after the preliminary check while the buffered input channel is full; the send then blocks forever before the new reply/close select can run. Make both sends select on w.input and w.closeCh. Also, TestWindowsCloseAddRace closes done itself before later selecting on done, so that select cannot detect whether Close completed, and Add has no timeout. Please make the test wait for a separate Close completion signal and bound the Add call. Review performed with OpenAI Codex at the account owners request.

